PR #23, opened 2017-07-21
huge cleanup

spacejam:
* replace rustc_serialize with idiomatic serde
* fully rip all b-tree logic out of the pagecache, so we can support other trees and clients
* rename `Pages` -> `PageCache`
* introduce the `PageMaterializer` trait for interfacing with the pagecache from other kinds of trees etc...
* nuke some unused tests


spacejam:
@pmuens yeah!

PR #24, opened 2017-07-22
Fix "unreachable pattern" error

pmuens:
Pattern matching captures all the different methods available for Frag. Therefore the catchall won't be triggered which results in a build error.


spacejam:
👍

PR #25, opened 2017-07-22
WIP - Update documentation

pmuens:
Uses Rust doc formatting so that the code is commented and documented at the same time and docs can be easily generated.

**NOTE:** Some of the doc comments are vague and some might be incorrect 🙈. Still in the progress of getting familiar with the codebase.

### DIFF
--- a/src/tree.rs
+++ b/src/tree.rs
@@ -8,6 +8,7 @@ use super::*;
 
 const FANOUT: usize = 2;
 
+/// Tree implementation which consists of `Page`s.
 #[derive(Clone)]
 pub struct Tree {
     pages: Arc<PageCache<BLinkMaterializer>>,
@@ -15,6 +16,7 @@ pub struct Tree {
 }
 
 impl Tree {
+    /// Create and setup a new `Tree`.
     pub fn new() -> Tree {
         let pages = PageCache::new(BLinkMaterializer, None);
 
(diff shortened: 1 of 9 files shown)


pmuens:
Tests are failing due to wrong example code in docs. Looking into it tomorrow morning!

pmuens:
Temporarily removed the examples from the docs (Amend the old commit). Will re-add them later on!

src/tree.rs
```diff
@@ -8,13 +8,15 @@ use super::*;
 
 const FANOUT: usize = 2;
 
+/// Tree implementation which consists of `Page`s.
```
spacejam: Lock-free `Tree` implementation based on a B-Link design (Lehman 81).

src/tree.rs
```diff
@@ -437,13 +450,15 @@ pub enum Data {
 }
 
 impl Data {
+    /// Length of the data stored in `Tree`.
```
spacejam: stored in the `Page`

src/tree.rs
```diff
@@ -475,7 +491,7 @@ impl Data {
     }
 }
 
-
+/// `Tree` `Node` which stores the actual `Data` and metadata about itself and nearby `Node`s.
```
spacejam: `Tree` `Node` which is either an index or leaf, and has a pointer to the right sibling if there is one.

spacejam:
closing for now, since a bunch of stuff has changed, but in general more docs are always good!

pmuens:
> closing for now, since a bunch of stuff has changed, but in general more docs are always good!

Yes, I agree. this one got quite stale and you pushed a couple of doc updates the other day. However thanks again for taking the time to look over this 👍

I'll add docs if I see that something is missing while working on the corresponding code.

PR #26, opened 2017-07-23
Add tests for DB

pmuens:
Add test which check the basic functionality of `DB`.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -31,3 +31,22 @@ impl DB {
         self.tree.del(&*k);
     }
 }
+
+#[test]
+fn basic_functionality() {
+    let db = DB::open();
+    let esl = &db.esl;
+    assert_eq!(esl.load(Ordering::SeqCst), 0);
+
+    db.insert(vec![1], vec![42]);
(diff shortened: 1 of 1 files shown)


spacejam:
I'm not sure if we want to think very hard about ESL yet, since it needs to be threaded into all operations through lower abstractions to coordinate higher-level transactions on multiple keys.

spacejam:
The `DB` module is pretty much just a placeholder at this point, and everything is likely to change, so it may not be worth having any tests for it yet, which may only break when we intentionally change the lower levels, which are already thoroughly tested.

The sorts of tests that will be eventually really useful for DB involve transactions on multiple keys, but we've got a bit of work to go before those are supported.

pmuens:
Thanks for the review @spacejam 👍

Gotcha. Yes, that makes sense. Was not aware that the `DB` implementation is just a placeholder for now.

(Wanted to jump into coding, that's why I thought adding tests could be a good start 😄).

Will look into the issues / roadmap to find smth. I can work on (maybe epoch based GC) 👍

PR #27, opened 2017-07-25
Log rework

spacejam:
Sorry for the monster PR, I think this is the last one of these for a while, since now I'm sort of happy with basic housekeeping.

This PR took a while because I introduced several bugs while refactoring the lock-free parts of the log, but in the end I think I fixed a lot more than I introduced (we'll see!). This stuff is really challenging to debug, because it's not deterministic, and it's really hard to reason about. I've been reminded many times over the last 3 days that you should only write something at 50% cleverness, because debugging is so much harder. There is a LOT of cleverness in here, which really hurt while hunting these bugs. But for now it seems to work, more or less :P We'll whip out the serious correctness tooling soon enough.

summary of changes:
* no more writer threadpool. we can re-add this in the future if we want the complexity. for now we don't need to think about shutdown logic or anything, and there are fewer tunables, although more latency bumps when it's time to write a buffer.
* rip out the crossbeam dependency, as it was only being used for feeding the writer threads
* rip out a random benchmarking binary that wasn't being used anymore
* move all log tests to the tests/ subdirectory
* generally rip out lots of excess Arc's. Since we're not handling multithreading internally, we can just force the user to use one when they share things across their threads now.

Looking forward to making more forward progress now that things are in a slightly cleaner state!


tests/test_tree.rs
```diff
@@ -49,7 +50,7 @@ fn kv(i: usize) -> Vec<u8> {
 #[test]
 fn it_works() {
     println!("========== initial sets ==========");
-    let t = Tree::new();
+    let t = Arc::new(Tree::new());
```
pmuens: Not sure if that's the root cause, but it looks like this change causes OS X builds to fail.

src/page.rs
```diff
@@ -1,17 +1,16 @@
...
-use serde::{Serialize, Deserialize};
+use bincode::{Infinite, serialize};
+use serde::{Deserialize, Serialize};
 
 use super::*;
 
 const MAX_FRAG_LEN: usize = 7;
 
-pub trait PageMaterializer {
+pub trait PageMaterializer: Send + Sync {
```
pmuens: Quick question about this addition: Is this part of a trait bound?

tests/test_tree.rs
spacejam: I think it's because I'm trying to open a file called `/dev/shm/...` which on linux machines actually just puts stuff into shared memory, but it's not present on OSX, so I'm going to use a trick described to me as "poor man's shared memory" where you create a file, then rm it when a process still has a reference open on it, which causes the pages to still exist in memory, but the writes never end up backed on disk.

spacejam: (shoutout to @alexginzburg for teaching me about that years ago :)

src/page.rs
spacejam: yeah, it means that anything that implements `PageMaterializer` must also implement `Send` and `Sync`

pmuens: Gotcha! That makes sense. 👍

tests/test_tree.rs
pmuens: Interesting 🤔!

Haven't seen the file usage. The "poor man's shared memory" approach is also new to me! Thanks for sharing 👍

PR #29, opened 2017-07-26
Rework modularization

### LINKED ISSUE #28: modularize core components
Right now some of the core components such as `tree`, `log`, etc. are larger files which include a couple of different `structs`, `enums` and their corresponding methods and functions.

Those components should be modularized so that a clear file hierarchy is present which makes it easier to understand and reason about the code, its different components and their relationships.

@spacejam already started the process for the `Log` component in https://github.com/spacejam/rsdb/pull/27.

pmuens:
Did a quick refactoring of the code structure so that it's modularized and therefore easier to grok.

Closes #28


src/bound/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod bound;
+
+pub use self::bound::*;
```
spacejam: usually you can move `<name>.rs` to `<name>/mod.rs` and it will feel a little more idiomatic.

For small stuff like bound, crc16, ops, radix, stack, I don't think there's much use splitting them into individual modules.

spacejam: so, I think really just tree and db should be split up for now. also feel free to rename PageMaterializer to just Materializer because it's namespaced to page.rs already.

spacejam: db is basically nonexistant right now, but will be a bit more complex after we implement recovery from log on startup, pagecache LRU, and epoch gc.

pmuens: Thanks for the review and feedback.

Yes, that makes sense 🤔. I was also not too impressed after taking another look at it this morning.

Having everything in separate directories when there is only one module makes it harder to reason about the code rather than easier.

I'll rebase this PR and update it according to your feedback :+1:

pmuens:
@spacejam just rebased and updated the code according to your PR review.

Just ping here if there's anything else to do 👍

pmuens:
Seems like #32 produced some conflicts here. Will look into it and resolve them!

pmuens:
Closing this one since I'll re-open another iteration of this in a new PR (re-implementing it real quick was easier than getting through the rebase pain).

pmuens:
New PR is #37

PR #37, opened 2017-07-30
Rework modularization

### LINKED ISSUE #28: modularize core components

pmuens:
Re-implementation of #29.

Closes #28


spacejam:
Looks great! Nice work!

pmuens:
Nice! Thanks for reviewing and merging @spacejam 👍 (Next up GC!)

PR #40, opened 2017-08-05
log speedup

spacejam:
~40% perf improvement by assigning the next log file offset at the time of sealing a buffer

cleanup of remaining code relying on hardcoded constants

some debugging improvements, log support (you won't see these log messages unless you call `env_logger::init();` in your code, but adding these messages makes seeing info easier in the future when debugging)

### DIFF
--- a/src/log/iobuf.rs
+++ b/src/log/iobuf.rs
@@ -12,6 +12,18 @@ struct IOBuf {
     log_offset: AtomicUsize,
 }
 
+impl Debug for IOBuf {
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        let header = self.get_header();
+        formatter.write_fmt(format_args!("\n\tIOBuf {{ log_offset: {}, n_writers: {}, offset: \
+                                          {}, sealed: {} }}",
+                                         self.get_log_offset(),
+                                         n_writers(header),
+                                         offset(header),
+                                         is_sealed(header)))
+    }
+}
+
 impl IOBuf {
     fn new(buf_size: usize) -> IOBuf {
         IOBuf {
@@ -58,7 +70,7 @@ impl IOBuf {
 
 pub struct IOBufs {
     bufs: Vec<IOBuf>,
-    current_buf: AtomicUsize,
+    sealed_bufs: AtomicUsize,
     written_bufs: AtomicUsize,
 
     intervals: Mutex<Vec<(LogID, LogID)>>,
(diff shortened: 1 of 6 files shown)


src/log/iobuf.rs
```diff
@@ -164,30 +161,40 @@ impl IOBufs {
...
+        assert!(buf.len() <= self.config.get_io_buf_size());
+
+        let mut printed = false;
+        macro_rules! trace_once {
+            ($($msg:expr),*) => {
+                if !printed {
+                    trace!($($msg),*);
+                    printed = true;
+                }};
+        }
```
spacejam: I wanted to use a macro here because I wanted to avoid writing `if !printed...` for each call to trace below, and macros let me pass variable numbers of arguments in, so I can retain the semantics of the `trace!` macro.

src/log/iobuf.rs
```diff
@@ -366,34 +402,13 @@ impl IOBufs {
...
-                spins = 0;
-            }
-            if self.bufs[(idx + 1) % N_BUFS].cas_log_offset(max, next_offset).is_ok() {
-                // success, now we can stop stalling and do other work
-                break;
-            }
-        }
-
-        // NB allows new threads to start writing into this buffer
-        iobuf.set_header(0);
```
spacejam: this was a big performance problem because we were spinning in the call to `reserve` until the log offset was set for a buffer that we wanted to use in a reservation, but it wasn't being set until the last thread using the previous io buffer finished their operation and wrote the data to disk. this meant that as soon as an io buffer was sealed, there was essentially a global mutex being held, preventing new reservations from being made until possibly multiple threads finished their work in the last IO buffer.

spacejam:
@pmuens  BTW, the way that I decided to spend time here was by using the script in `hack/flamerun.sh` like this:
```
# compile the test code
cargo test

# run the test binary and generate a performance flamegraph
hack/flamerun.sh ./target/debug/test_log-347107ad085f1906

# open the resulting file in an SVG viewer (ff)
firefox flamegraph.svg
```

the resulting svg shows where different threads spend their time during the tests, and you can click on a particular section to zoom in. I saw that a lot of time was being spent calling `get_log_offset` during the reservation process, which happens after claiming a buffer and WAS done in a loop that breaks whenever the associated log offset for the buffer is actually set, which could take some time due to the reasons I wrote about earlier.

pmuens:
> @pmuens BTW, the way that I decided to spend time here was by using the script in hack/flamerun.sh

Great! That's good to know. Thanks for creating and sharing @spacejam 👍

PR #41, opened 2017-08-05
Add docs to all exported things.

spacejam:
I wanted to get the docs into a better spot, so I just filled them in.


src/lib.rs
```diff
@@ -1,3 +1,7 @@
+//! `rsdb` is a flash-sympathetic persistent lock-free B+ tree.
+#![cfg_attr(test, deny(warnings))]
+#![deny(missing_docs)]
```
spacejam: These lines will enforce slightly better coding style.

PR #44, opened 2017-08-07
port the rocksdb hash function to rust

spacejam:


spacejam:
this will be used for determining a cache shard very frequently

PR #45, opened 2017-08-07
compression. closes #7.

### LINKED ISSUE #7: compression of items written to log slots
we can't compress on the log-buffer level without destroying the property of a `LogID` being its physical offset, so we need to compress before writing here. Desired properties:
* works well with small amounts of bytes of human languages
* configurable

spacejam:


spacejam:
uses facebook's zstd compression library, which has a really low cpu cost considering the nice compression ratios its able to achieve.

src/log/iobuf.rs
```diff
@@ -431,17 +437,23 @@ impl Drop for IOBufs {
 }
 
 #[inline(always)]
-fn encapsulate(buf: &mut Vec<u8>) -> Vec<u8> {
+fn encapsulate(raw_buf: Vec<u8>, use_compression: bool) -> Vec<u8> {
+    let mut buf = if use_compression {
+        compress(&*raw_buf, 5).unwrap()
```
spacejam: actual compression happens here

src/log/lss.rs
```diff
@@ -18,62 +20,81 @@ impl LockFreeLog {
...
             let msg = format!("read data failed crc16 checksum, {:?} should be {:?}",
                               checksum,
                               crc16_buf);
-            return Err(Error::new(ErrorKind::Other, msg));
+            return (Err(Error::new(ErrorKind::Other, msg)), len);
         }
 
-        Ok(Ok(buf))
+        if self.config().get_use_compression() {
+            (Ok(Ok(decompress(&*buf, max).unwrap())), len)
```
spacejam: actual decompression happens here

src/log/lss.rs
```diff
@@ -18,62 +20,81 @@ impl LockFreeLog {
...
-        self.iobufs.config()
-    }
-
-    fn write(&self, buf: Vec<u8>) -> LogID {
-        self.iobufs.reserve(buf).complete()
-    }
-
-    /// read a buffer from the disk
-    fn read(&self, id: LogID) -> io::Result<Result<Vec<u8>, usize>> {
+    fn read_with_raw_sz(&self, id: LogID) -> (io::Result<Result<Vec<u8>, usize>>, usize) {
```
spacejam: needed this internal function for returning the raw length as well as the possible item, for use with the iterator. before, the iterator was advancing the next offset by the length of the message it read from the log, but when using compression this messes things up, and we need to advance a (usually much lower) number depending on the number of compressed bytes on disk

src/config.rs
```diff
@@ -138,6 +221,12 @@ impl Config {
                 options.read(true);
                 options.write(true);
 
+                if !self.use_os_cache && cfg!(unix) {
+                    use std::os::unix::fs::OpenOptionsExt;
+                    options.custom_flags(libc::O_DIRECT);
+                    panic!("O_DIRECT support not sussed out yet.");
```
spacejam: this will eventually let us completely bypass the OS cache. when using compression, the OS cache just stores compressed disk blocks, which is less useful to us than the decompressed part, so sometimes you want to disable it.

PR #47, opened 2017-08-09
periodically flush dirty data, page-out when cache is too full. closes #3. closes #4.

### LINKED ISSUE #3: dirty page flushing
- [x] periodically flush the dirty buffer
- [x] make the `LogID` in `CacheEntry` non-optional

in the future we may want to buffer things before claiming a log slot for updates, but for now let's just keep it simple

### LINKED ISSUE #4: multi-fragment flush strategy
to improve data locality and reduce written data, write adjacent chunks of frags in a single stack at once to the log

spacejam:


src/log/iobuf.rs
```diff
@@ -13,6 +13,8 @@ struct IOBuf {
     log_offset: AtomicUsize,
 }
 
+unsafe impl Sync for IOBuf {}
```
spacejam: this basically tells the compiler "this is responsible for its own synchronization if it's shared by multiple threads"

It was necessary to add it now because in the periodic flusher, we send an `Arc` containing an `IOBufs` to another thread. We didn't need to specify this before, even though we have multithreaded tests, because `Sync` is implemented for higher-level structures that contain `IOBufs` already, and if you mark a container as `Sync`, it doesn't require that all members are recursively `Sync` as well. This is part of why it's `unsafe`, because you're basically telling the compiler "don't worry, I've got this".

This can lead to really subtle bugs, even in the standard library. This article,  https://www.ralfj.de/blog/2017/06/09/mutexguard-sync.html talks about how `MutexGuard` (the thing returned by calling `lock()` on a mutex, which implements `Deref` and `DerefMut` to pass accesses to the underlying memory contained in the `Mutex`) used to implement  `Sync`. This is bad, because if something is `Sync`, it can be sent to multiple threads simultaneously. This let you unlock a mutex, then send the result to multiple threads, even though the whole point of a mutex is to make sure multiple threads can't access something simultaneously.

src/log/lss.rs
```diff
@@ -10,18 +12,42 @@ use super::*;
...
 }
 
 unsafe impl Send for LockFreeLog {}
 unsafe impl Sync for LockFreeLog {}
 
+impl Drop for LockFreeLog {
+    fn drop(&mut self) {
+        self.flusher_shutdown.store(true, std::sync::atomic::Ordering::SeqCst);
+        if let Some(join_handle) = self.flusher_handle.take() {
+            join_handle.join().unwrap();
```
spacejam: we call `join()` here to wait until the flusher thread wakes up, reads the flusher_shutdown atomic bool (now set to true) and returns. we need to do this because the flusher thread still has a reference to the underlying `IOBufs`. This is important because in the impl of `Drop` for `IOBufs`, it calls `self.flush()` for each of its buffers, making sure all pending data is written to disk. If we return from this destructor without making sure that the flusher thread has finished (dropping the Arc count to 1, for the `LockFreeLog` which has the other reference) then the data may not be flushed by the time the process quits. This basically forces flushing of pending writes before the `LockFreeLog` itself is dropped.

This was causing a panic before I did this, in  `tests/test_pagecache.rs (basic_recovery test)`  because the old log was being dropped before data was flushed, then a new pagecache was created, and it tried to read a page that should have been present, but it hadn't been flushed yet, and its test failed.

src/log/periodic_flusher.rs
```diff
@@ -0,0 +1,22 @@
...
+               -> std::io::Result<thread::JoinHandle<()>> {
+    thread::Builder::new()
+        .name(name)
+        .spawn(move || {
+            while !shutdown.load(SeqCst) {
+                iob.flush();
+                thread::sleep(Duration::from_millis(flush_every_ms));
+            }
+        })
+}
```
spacejam: this is the entire flusher thread. it just spins and flushes every `flush_every_ms`, until something changes the atomic bool that it watches to `true`, signalling that it should shutdown.

src/page/lru.rs
```diff
@@ -18,139 +15,130 @@ impl Lru {
...
-        let mut shard = shard_mu.write().unwrap();
+        let mut shard = shard_mu.lock().unwrap();
+        shard.page_out_succeeded(pid);
+    }
 
-        shard.bump(lid);
+    fn idx(&self, pid: PageID) -> usize {
+        let k: [u8; 8] = unsafe { std::mem::transmute(pid) };
+        let h = hash::hash(&k, 0) as usize;
+        h % (2 << self.cache_bits)
```
spacejam: this function turns an 8 byte long `PageID` into an array of 8 bytes, then passes that array to the hash function, then uses modulus on the cache size (we have 2 ^ cache_bits total shards).

src/page/lru.rs
```diff
@@ -18,139 +15,130 @@ impl Lru {
...
 
-    fn pop(&mut self, lid: LogID) -> Result<(), ()> {
-        let entry_opt = self.entries.remove(&lid);
-        if entry_opt.is_none() {
-            return Err(());
-        }
+        entry.accesses += 1;
+        entry.sz = sz;
+        // guaranteed to be higher than any other time measured before
+        entry.mtime = self.ctime.elapsed();
```
spacejam: by calling `elapsed` here, we are guaranteed to get a higher number than has ever been returned before for this `Shard`. This is because the result is guaranteed to be monotonic. So, we don't need to worry about overwriting a different cache entry when we insert the entry into the `self.accesses` `BTreeMap`

src/page/page_cache.rs
```diff
@@ -179,54 +178,102 @@ impl<M> PageCache<LockFreeLog, M>
...
+                    let (pp, lid) = match *ce {
+                        &mut CacheEntry::Resident(ref pp, ref lid) => (pp.clone(), lid.clone()),
+                        &mut CacheEntry::Flush(lid) => (self.pull(lid).unwrap(), lid),
+                        &mut CacheEntry::PartialFlush(lid) => (self.pull(lid).unwrap(), lid),
+                    };
+                    **ce = CacheEntry::Resident(pp, lid);
+                });
+
+            let node = node_from_frag_vec(cache_entries.clone());
+            let res = unsafe { (*stack_ptr).cas(head, node) };
```
spacejam: here's the bug I realized existed while we were talking, I had forgotten to actually swap the stack after reading the values off of disk. Now it only reads these off disk when they are not actually present (`contains_non_resident` above)

src/page/page_cache.rs
```diff
@@ -235,7 +282,9 @@ impl<M> PageCache<LockFreeLog, M>
         partial_pages.reverse();
         let partial_pages = partial_pages;
 
-        let mut head = head;
+        let to_evict = self.lru.accessed(pid, std::mem::size_of_val(&partial_pages));
+        self.page_out(to_evict);
```
spacejam: the lru may tell us at any time that we need to clear-up some space, then we use `page_out` to clean up memory.

PR #60, opened 2017-08-10
add clippy

spacejam:
I've added the clippy tool, which will now cause build errors when it's not happy. Clippy will run with `cargo test` but not `cargo build`.


spacejam:
basic changes:
* tweak .travis.yml to conditionally use clippy, and use cache
* address tons of little clippy lints
* improve documentation for the crate, in `src/lib.rs`

PR #63, opened 2017-08-11
add a macro for making configuration options slightly less error-prone

spacejam:


src/config.rs
```diff
@@ -37,188 +39,55 @@ impl Default for Config {
...
-    pub fn cache_capacity(&self, cache_capacity: usize) -> Config {
-        let mut ret = self.clone();
-        ret.cache_capacity = cache_capacity;
-        ret
-    }
+macro_rules! builder {
+    ($(($name:ident, $get:ident, $set:ident, $t:ty, $desc:expr)),*) => {
+        $(
+            #[doc="Get "]
+            #[doc=$desc]
```
spacejam: the macro system is sort of black magic, and pretty annoying, but in cases like this it can reduce tedium significantly.

src/config.rs
```diff
@@ -37,188 +39,55 @@ impl Default for Config {
...
+            }
 
-    /// Builder, set whether the system compresses data written to
-    /// secondary storage with the zstd algorithm.
-    pub fn use_compression(&self, use_compression: bool) -> Config {
-        let mut ret = self.clone();
-        ret.use_compression = use_compression;
-        ret
+            #[doc="Builder, set "]
+            #[doc=$desc]
```
spacejam: I have 2 `doc` attributes because I don't know of a way to concatenate things like this in a macro.

src/config.rs
```diff
@@ -37,188 +39,55 @@ impl Default for Config {
...
-    }
-
-    /// Builder, set the cache capacity in bytes.
-    pub fn cache_capacity(&self, cache_capacity: usize) -> Config {
-        let mut ret = self.clone();
-        ret.cache_capacity = cache_capacity;
-        ret
-    }
+macro_rules! builder {
+    ($(($name:ident, $get:ident, $set:ident, $t:ty, $desc:expr)),*) => {
```
spacejam: this like basically expands to meaning: "there will be groups of 5 things (3 identifiers, 1 type name, and an expression) within sets of parentheses, and after the parentheses, there will be a comma if there is another group afterward"

the `$(a)b*` pattern means, `a` may be repeated, separated by `b`. in this case.

src/config.rs
```diff
@@ -37,188 +39,55 @@ impl Default for Config {
...
+        $(
```
spacejam: this beginning `$(` in the macro body means that this is the section that will be repeated for each match above

src/config.rs
```diff
@@ -37,188 +39,55 @@ impl Default for Config {
...
-        ret.use_compression = use_compression;
-        ret
+            #[doc="Builder, set "]
+            #[doc=$desc]
+            pub fn $name(&self, to: $t) -> Config {
+                let mut ret = self.clone();
+                ret.$name = to;
+                ret
+            }
+        )*
```
spacejam: this `)*` means the repeated block will end here

spacejam: it's also possible to add a custom separator before the `*` if necessary, but it's not here

src/config.rs
```diff
@@ -37,188 +39,55 @@ impl Default for Config {
...
+        (page_consolidation_threshold, get_page_consolidation_threshold, set_page_consolidation_threshold, usize, "page consolidation threshold"),
+        (path, get_path, set_path, Option<String>, "path for the main storage file"),
+        (cache_bits, get_cache_bits, set_cache_bits, usize, "log base 2 of the number of cache shards"),
+        (cache_capacity, get_cache_capacity, set_cache_capacity, usize, "maximum size for the system page cache"),
+        (use_os_cache, get_use_os_cache, set_use_os_cache, bool, "whether to use the OS page cache"),
+        (use_compression, get_use_compression, set_use_compression, bool, "whether to use zstd compression"),
+        (flush_every_ms, get_flush_every_ms, set_flush_every_ms, Option<usize>, "number of ms between IO buffer flushes"),
+        (snapshot_after_ops, get_snapshot_after_ops, set_snapshot_after_ops, usize, "number of operations between page table snapshots"),
+        (snapshot_path_prefix, get_snapshot_path_prefix, set_snapshot_path_prefix, Option<String>, "snapshot file prefix")
+    );
```
spacejam: and the result is a super-compressed, faster way to add new configuration arguments!

spacejam: it's sort of weird that you have to say 3 names, instead of the macro being able to "concatenate identifiers" like just putting "set_" and "get_" in front of the first name. this is waiting on a feature called "concat_idents!" which I don't think has gotten added yet.

Issue #56, opened 2017-08-09
implement mvp benchmark calls to db


spacejam:
take the arguments, and implement these operations:
1. spin up the specified number of threads
1. execute the desired number of total operations
1. roughly maintain the specified proportion of reads, writes, cas, del, and scan operations among the operations

pmuens:
Quick update on this one that the conversation in #55 should be taken into account here.

Specifically:

- https://github.com/spacejam/rsdb/issues/55#issuecomment-321545616
- https://github.com/spacejam/rsdb/issues/55#issuecomment-322003016

PR #66, opened 2017-08-11
Read CLI args for benchmarking tool

### LINKED ISSUE #55: read in arguments for benchmark tool
the benchmark tool should accept arguments for these parameters:
1. number of threads
1. number of total operations
1. proportion of reads, writes, cas, del, and scan operations among the operations
1. freshness bias (prefer recent/likely to be in cache, prefer old/not likely to be in cache, no preference)
1. non-present-key chance (the chance that a request for cas/get/set/del may be sent for a key that does not exist)
1. key size min, max, median
1. value size min, max, median
1. scan iterations min, max, median

pmuens:
Closes #55 

Can be tested via `cargo build && ./target/debug/bench --num-threads 42`

### DIFF
--- a/src/bin/bench.rs
+++ b/src/bin/bench.rs
@@ -1,0 +1,489 @@
+extern crate clap;
+extern crate num_cpus;
+extern crate rsdb;
+#[macro_use]
+extern crate log;
+
+use std::error::Error;
+use std::io::prelude::*;
+use std::process;
+use std::collections::HashMap;
+use clap::{App, Arg};
+
+fn main() {
+    let cpus = &(num_cpus::get().to_string());
+
+    let matches = App::new("RSDB bench")
+        .version("0.1.0")
+        .about("RSDB benchmarking tool")
+        .author("Tyler Neely <t@jujit.su>")
+        .arg(Arg::with_name("num_threads")
+            .short("t")
+            .long("num-threads")
+            .help("Number of threads to use")
+            .default_value(cpus)
+            .takes_value(true))
+        .arg(Arg::with_name("num_operations")
+            .short("o")
+            .long("num-operations")
+            .help("Number of total operations")
+            .default_value("1000000")
+            .takes_value(true))
+        .arg(Arg::with_name("freshness_bias")
+            .long("freshness-bias")
+            .help("Freshness bias")
+            .possible_value("old")
+            .possible_value("new")
+            .possible_value("random")
+            .default_value("random")
+            .takes_value(true))
+        .arg(Arg::with_name("non_present_key_chance")
+            .long("non-present-key-chance")
+            .help("The chance that a request may be sent for a key that does not exist")
+            .default_value("true")
+            .takes_value(true))
+        // proportions
+        .arg(Arg::with_name("proportion_set")
+            .long("prop-set")
+            .help("Proportion of sets")
+            .default_value("10")
+            .takes_value(true))
+        .arg(Arg::with_name("proportion_scan")
+            .long("prop-scan")
+            .help("Proportion of scan")
+            .default_value("4")
+            .takes_value(true))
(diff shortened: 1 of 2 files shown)


src/bin/bench.rs
```diff
@@ -0,0 +1,489 @@
...
+            scan_iter_min: scan_iter_min.unwrap(),
+            scan_iter_max: scan_iter_max.unwrap(),
+            scan_iter_median: scan_iter_median.unwrap(),
+        })
+    }
+}
+
+fn check_min_max_median(min: u64, max: u64, median: u64) {
+    if !(min <= median && median <= max) {
+        warn!("Please check and provide different {}, {} and {} values", min, max, median);
```
spacejam: "Please check and provide different min({}), max({}) and median({}) values"

src/bin/bench.rs
```diff
@@ -0,0 +1,489 @@
...
+extern crate num_cpus;
+extern crate rsdb;
+#[macro_use]
+extern crate log;
+
+use std::error::Error;
+use std::io::prelude::*;
+use std::process;
+use std::collections::HashMap;
+use clap::{App, Arg};
```
spacejam: this is sort of pedantic, but I usually like to separate the standard library and external imports into different blocks, to help me see off the bat which things are external

pmuens: I love that. Will update it so that it looks like the rest of the codebase.

spacejam:
nice work! maybe we can rename the options that start with `prop` to just the part after? like `prop-cas` to just `cas`? I didn't think about this when reading the code or your proposal before, but running the command in the terminal makes it feel like maybe we can trim it down.

pmuens:
> nice work!

Thanks for the kind words and the review @spacejam 👍

> maybe we can rename the options that start with prop to just the part after? like prop-cas to just cas? I didn't think about this when reading the code or your proposal before, but running the command in the terminal makes it feel like maybe we can trim it down.

Yes, that sounds like a good idea. When thinking about it I associate `prop` with `property` rather than `proportion`!

I'll update the PR tomorrow 👍

pmuens:
@spacejam just pushed some updates and fixed the different things you mentioned in the review yesterday.

Next up I'll work on the other requirements you layed out in https://github.com/spacejam/rsdb/issues/55#issuecomment-321545616.

pmuens:
@spacejam I just pushed some changes and resolved the comments from your PR review 👍

I'd say that this covers the todos in #55. The conversation we had at the end of #55 is more related to #56 and IMHO could be implemented in a new PR which will close #56 (Will jump into this next up).

Any objections which prevents this from getting merged? Should I fix anything else?

src/bin/bench.rs
```diff
@@ -0,0 +1,503 @@
...
+
+use clap::{App, Arg};
+
+fn main() {
+    let cpus = &(num_cpus::get().to_string());
+
+    let matches = App::new("RSDB bench")
+        .version("0.1.0")
+        .about("RSDB benchmarking tool")
+        .author("Tyler Neely <t@jujit.su>")
```
spacejam: this should be your name, since you're building this tool!

pmuens: 😅 thanks (don't wanted to be rude here)!

Just pushed a change with an update! 👍

spacejam:
LGTM! If in the future you want to reduce the code that is basically there to put stuff into the Config struct, I found this argument parsing crate that simplifies the process a LOT: https://github.com/docopt/docopt.rs/blob/master/examples/decode.rs

but since this does the same thing, no need to do more work to replicate existing functionality, unless the amount of code gets on your nerves.

pmuens:
Thanks for merging @spacejam 👍

That crate looks sweet! I'll look into it to see if and how it can be used in conjunction with Clap.

Next up I'll look into #56.

PR #68, opened 2017-08-14
periodically snapshot pagetable, recover from snapshots. closes #9. closes #21.

### LINKED ISSUE #9: log replay after restoring page table snapshot

### LINKED ISSUE #21: pagetable snapshotting
on recovery, we should be able to recover a snapshot of the pagetable, then read the log after the pagetable was created for full reconstruction. we only want to read and recover the structure of pages and the offsets of chunks of frags in the log for future retrieval, without pulling in any kv data.

initially this can be done by forking and snapshotting, similar to redis

spacejam:


src/page/mod.rs
```diff
@@ -45,41 +46,62 @@ pub trait Materializer: Send + Sync + Clone {
...
 }
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+struct Snapshot<R> {
+    pub max_lid: LogID,
+    pub max_pid: PageID,
+    pub pt: BTreeMap<PageID, Vec<LogID>>,
+    pub free: Vec<PageID>,
+    pub recovery: Option<R>,
+}
```
spacejam: This is the main structure that gets written to disk when we take a snapshot. Note that we store only a Vec of LogID's, which is a pointer to where to find the data later, rather than any actual data.

src/tree/mod.rs
```diff
@@ -406,7 +407,7 @@ impl Tree {
             true,
         );
         self.pages
-            .prepend(new_root_pid, new_root_cas_key, new_root)
+            .replace(new_root_pid, new_root_cas_key, vec![new_root])
```
spacejam: replace is used as a signifier that a new page is being written over a previous one completely, or it's the first time we're using a page

src/tree/mod.rs
```diff
@@ -442,6 +443,11 @@ impl Tree {
     }
 
     #[doc(hidden)]
+    pub fn __delete_all_files(self) {
+        self.pages.__delete_all_files();
+    }
```
spacejam: this is mostly a test helper, hence the `#[doc(hidden)]`

src/tree/mod.rs
```diff
@@ -630,14 +636,15 @@ impl Materializer for BLinkMaterializer {
         vec![Frag::Base(base_node, root)]
     }
 
-    fn recover(&mut self, frag: &Frag) -> Option<PageID> {
+    fn recover(&self, frag: &Frag) -> Option<PageID> {
```
spacejam: recover is now responsible for providing its own interior mutability, because it may be called by any thread that is performing a snapshot, as it "catches up" with the new log tip, from the last snapshot.

spacejam: (this is why a Mutex was added)

src/page/page_cache.rs
```diff
@@ -1,50 +1,65 @@
...
-    max_id: AtomicUsize,
+pub struct PageCache<PM, L, P, R> {
+    t: PM,
+    inner: Radix<Stack<CacheEntry<P>>>,
+    max_pid: AtomicUsize,
     free: Stack<PageID>,
-    log: Box<L>,
+    log: L,
     lru: lru::Lru,
+    updates: AtomicUsize,
```
spacejam: updates tracks the number of operations, checked against the configurable number of updates between snapshots

src/page/page_cache.rs
```diff
@@ -1,50 +1,65 @@
...
+    last_snapshot: AtomicOption<Snapshot<R>>,
```
spacejam: we store the last snapshot so that when we generate a new one, we can create a diff by scanning the log between the last snapshot point and the tip.

src/page/page_cache.rs
```diff
@@ -1,50 +1,65 @@
...
+}
+
+unsafe impl<PM, L, P, R> Send for PageCache<PM, L, P, R>
+    where PM: Send,
+          L: Send,
+          R: Send
+{
 }
```
spacejam: we only want to send this across threads if it's OK to send the page materializer, log, and recovery types, too

src/page/page_cache.rs
```diff
@@ -1,50 +1,65 @@
...
 
-unsafe impl<L: Log, M: Materializer> Send for PageCache<L, M> {}
-unsafe impl<L: Log, M: Materializer> Sync for PageCache<L, M> {}
+unsafe impl<PM, L, P, R> Sync for PageCache<PM, L, P, R>
+    where PM: Sync,
+          L: Sync
```
spacejam: we only want to share this with multiple threads if the page materializer and log are OK with that too

src/page/page_cache.rs
```diff
@@ -1,50 +1,65 @@
...
-impl<M> PageCache<LockFreeLog, M>
-    where M: Materializer,
-          M::PartialPage: Clone,
-          M: DeserializeOwned,
-          M: Serialize
+impl<PM, P, R> PageCache<PM, LockFreeLog, P, R>
+    where PM: Materializer<PartialPage = P, Recovery = R>,
+          PM: Send + Sync,
+          P: PartialEq + Clone + Debug + Serialize + DeserializeOwned + Send,
+          R: PartialEq + Clone + Debug + Serialize + DeserializeOwned
```
spacejam: I split the informatino about the Materializer's associated types because I was having trouble showing that the materializer's recovery type was deserializable before this. Sometimes splitting things out more lets the compiler see that something is actually possible.

src/page/page_cache.rs
```diff
@@ -1,50 +1,65 @@
...
     /// Instantiate a new `PageCache`.
-    pub fn new(pm: M, config: Config) -> PageCache<LockFreeLog, M> {
+    pub fn new(pm: PM, config: Config) -> PageCache<PM, LockFreeLog, P, R> {
         let cache_capacity = config.get_cache_capacity();
         let cache_shard_bits = config.get_cache_bits();
         PageCache {
             t: pm,
             inner: Radix::default(),
-            max_id: AtomicUsize::new(0),
+            max_pid: AtomicUsize::new(0),
```
spacejam: made this explicitly `max_pid` because I hit a bug when developing this feature where I confused `PageID`'s and `LogID`'s. This is pretty much the only case where golang has more compiler strictness, around typecasts. Rust treats them both as `u64`'s and you can mix them up easier.

src/page/page_cache.rs
```diff
@@ -109,8 +123,9 @@ impl<M> PageCache<LockFreeLog, M>
                         let stack = raw(Stack::default());
                         self.inner.insert(prepend.pid, stack).unwrap();
                         free_pids.retain(|&pid| pid != prepend.pid);
-                        if self.max_id.load(SeqCst) < prepend.pid {
-                            self.max_id.store(prepend.pid, SeqCst);
+                        if self.max_pid.load(SeqCst) <= prepend.pid {
+                            // always make sure self.max_pid is higher than what exists
+                            self.max_pid.store(prepend.pid + 1, SeqCst);
```
spacejam: add 1 to what we store because self.max_pid should always be higher than what is already present (this turned into a bug when I was writing this)

src/page/page_cache.rs
```diff
@@ -359,7 +395,211 @@ impl<M> PageCache<LockFreeLog, M>
...
+
+        let snapshot_opt = self.last_snapshot.take(SeqCst);
+        if snapshot_opt.is_none() {
+            // some other thread is snapshotting
+            warn!(
+                "snapshot skipped because previous attempt \
+                  appears not to have completed"
+            );
+            return;
+        }
```
spacejam: with the `AtomicOption`, we are enforcing the invariant that only one thread will ever be running a snapshot simultaneously. You need to `take()` it (which is atomic) and hold the old snapshot before you're allowed to create a new one.

This is more of a capacity-related invariant than anything, as it should be totally safe to create as many snapshots as you want, or delete them, etc... A snapshot is an immutable optimization over an immutable underlying log.

src/page/page_cache.rs
```diff
@@ -359,7 +395,211 @@ impl<M> PageCache<LockFreeLog, M>
...
+
+        let raw_bytes = serialize(&snapshot, Infinite).unwrap();
+        let bytes = if self.config().get_use_compression() {
+            compress(&*raw_bytes, 5).unwrap()
+        } else {
+            raw_bytes
+        };
+        let crc64: [u8; 8] = unsafe { std::mem::transmute(crc64::crc64(&*bytes)) };
+
+        self.last_snapshot.swap(snapshot, SeqCst);
```
spacejam: as soon as we populate the snapshot `AtomicOption`, another thread may start a new snapshot at any time

src/page/page_cache.rs
```diff
@@ -359,7 +395,211 @@ impl<M> PageCache<LockFreeLog, M>
...
+
+        // write the snapshot bytes, followed by a crc64 checksum at the end
+        f.write_all(&*bytes).unwrap();
+        f.write_all(&crc64).unwrap();
+        f.sync_all().unwrap();
+        drop(f);
+
+        if let Err(e) = std::fs::rename(path_1, &path_2) {
+            error!("failed to write snapshot: {}", e);
+        }
```
spacejam: Basic file write strategy:
1. open non-final path that signifies it's not finished with the name (`in___motion`)
1. write the snapshot and checksum
1. move (which is atomic) the file so that its name signifies that it's complete

src/page/page_cache.rs
```diff
@@ -359,7 +395,211 @@ impl<M> PageCache<LockFreeLog, M>
...
+        let prefix_glob = format!("{}.[0-9]*", prefix);
+        for entry in glob::glob(&*prefix_glob).unwrap() {
+            if let Ok(entry) = entry {
+                if entry.to_str().unwrap() != &*path_2 {
+                    info!("removing old snapshot file {:?}", entry);
+                    std::fs::remove_file(entry).unwrap();
+                }
+            }
+        }
+    }
```
spacejam: delete any old snapshots still hanging around

this is sort of buggy, since it's technically possible that another thread came in and wrote another snapshot after we `swap()`'d our snapshot into the `AtomicOption`, but it's pretty unlikely, and does not impact correctness.

src/page/page_cache.rs
```diff
@@ -359,7 +395,211 @@ impl<M> PageCache<LockFreeLog, M>
...
+                if !entry_pb.to_str().unwrap().contains(".in___motion") {
+                    candidates.push(entry_pb);
+                }
+            }
+        }
+        if candidates.is_empty() {
+            return Snapshot::default();
+        }
+
+        candidates.sort_by_key(|path| path.metadata().unwrap().created().unwrap());
```
spacejam: again, we are sort of buggy here, because we're just sorting the files by created time and picking the newest one, but it's probably OK, and doesn't impact overall system correctness, since this is all an optimization anyway.

src/page/mod.rs
pmuens: So `LogID` mean that the data was written to disk previously and we don't have pointer which might point to in-memory data, right?

spacejam: right, we need to go to those log offsets to read the data into memory if we need to read them

PR #71, opened 2017-08-14
use quickcheck to test basic tree operations, comparing to a BTreeMap. closes #50.

### LINKED ISSUE #50: quickcheck on tree

spacejam:


tests/test_tree.rs
```diff
@@ -162,4 +159,128 @@ fn recovery() {
...
+        let mut smaller = vec![];
+        for i in 0..self.ops.len() {
+            let mut clone = self.clone();
+            clone.ops.remove(i);
+            smaller.push(clone);
+        }
+
+        Box::new(smaller.into_iter())
+    }
+}
```
spacejam: the `shrink` function is used to make a failing test case more manageable to understand. it will generate a shorter sequence of operations, and keep dropping items out until it finds the shortest possible set of operations that trigger the violation of the property

tests/test_tree.rs
```diff
@@ -162,4 +159,128 @@ fn recovery() {
...
+
+        let choice = g.gen_range(0, 5);
+
+        match choice {
+            0 => Op::Set(g.gen::<u8>(), g.gen::<u8>()),
+            1 => Op::Get(g.gen::<u8>()),
+            2 => Op::Del(g.gen::<u8>()),
+            3 => Op::Cas(g.gen::<u8>(), g.gen::<u8>(), g.gen::<u8>()),
+            4 => Op::Scan(g.gen::<u8>(), g.gen_range::<usize>(0, 40)),
+            _ => panic!("impossible choice"),
```
spacejam: we generate a sequence of operations on the tree, hoping to trigger bugs. similar to fuzzing

tests/test_tree.rs
```diff
@@ -162,4 +159,128 @@ fn recovery() {
...
+                    .take(len)
+                    .map(|(ref rk, ref rv)| (**rk, **rv));
+                for (t, r) in tree_iter.zip(ref_iter) {
+                    assert_eq!(t, r);
+                }
+            }
+            Restart => {
+                drop(tree);
+                tree = config.tree();
+            }
```
spacejam: for each operation, we apply the same thing to the `BTreeMap` (except restart, which exercises the recovery logic). For things that return results, we ensure that both things return the same thing. If our tree returns something else, there's a bug somewhere.

spacejam:
it's already found a couple bugs, by tweaking some tree parameters :) this kind of test is super powerful, since it avoids the biases present in normal unit tests

pmuens:
> it's already found a couple bugs, by tweaking some tree parameters :) this kind of test is super powerful, since it avoids the biases present in normal unit tests

Nice! That sounds promising :+1: Love the approach of having more randomized inputs for tests :+1:

PR #72, opened 2017-08-14
MVP Bench calls

### LINKED ISSUE #56: implement mvp benchmark calls to db

pmuens:
Closes #56


src/bin/bench.rs
```diff
@@ -159,9 +165,88 @@ fn main() {
...
+    // create a default rsdb config
+    let rsdb_config = rsdb::Config::default();
+    let mut tree = rsdb_config.tree();
+
+    pool.install(|| perform_operations(&mut tree, config));
+
     Ok(())
 }
 
+fn perform_operations(tree: &mut Tree, config: Config) {
```
spacejam: does tree need to be `&mut` or should it just be `&`? Rayon will be passing a reference to the tree to multiple threads, so I think it needs to be an immutable reference. The tree handles mutability by using atomic operations, so it is actually made to be shared between multiple threads without a mutex or needing a mutable reference.

src/bin/bench.rs
```diff
@@ -159,9 +165,108 @@ fn main() {
...
+    for i in 0..config.num_operations {
+        let op = get_operation_choice(&ops, sum_ops);
+
+        let operation: Box<Fn() + Sync> = match op {
+            Some(&Op::Set) => Box::new(|| perform_set_operation(&tree)),
+            Some(&Op::Scan) => Box::new(|| perform_scan_operation(&tree)),
+            Some(&Op::Get) => Box::new(|| perform_get_operation(&tree)),
+            Some(&Op::Delete) => Box::new(|| perform_delete_operation(&tree)),
+            Some(&Op::Cas) => Box::new(|| perform_cas_operation(&tree)),
+            _ => Box::new(|| {}),
```
spacejam: one thing I worry about is if we have to do billions of operations, how much memory will that take up? it may be better to generate the operation on the fly rather than creating a vector of operations to pop off for each one, and probably faster too.

pmuens: Yes, you're completely right. I'll update it :+1:

Anyway. Let me explain why I did this in the first place. So the problem was that I thought that I should use the rayons `join` method to join all the closures. That's why I generated a vector of closures which I wanted to `join` later on.

However `join` only takes two operations and according to the docs on could also use `install` to run the operations in the thread pool. So this is basically legacy code I forgot to remove because I switched from `join` to `install` a few minutes before I pushed the code.

pmuens:
@spacejam just pushed some changes and updates.

This PR should implement everything defined in #56.

One thing I'm curious about is how we should deal with operations other than `set`. So right now we generate a random KV pair for each operation. This means that e.g. for `set` there is nearly a 0% chance that it will find the value which it requested (same for other operations).

Don't know if this should be by design?! A solution would be to generate a fixed set of KV pairs before preforming the tree operations. However the problem with that is that we don't know how many KV pairs are needed because we compute the `get_operation_choice` which will return the operation to perform on the fly 🤔.

src/bin/bench.rs
```diff
@@ -159,29 +165,151 @@ fn main() {
...
+        let op: Box<Fn() + Sync> = match op_to_perform {
+            Some(&Op::Set) => Box::new(|| perform_set_operation(&tree)),
+            Some(&Op::Scan) => Box::new(|| perform_scan_operation(&tree)),
+            Some(&Op::Get) => Box::new(|| perform_get_operation(&tree)),
+            Some(&Op::Delete) => Box::new(|| perform_delete_operation(&tree)),
+            Some(&Op::Cas) => Box::new(|| perform_cas_operation(&tree)),
+            _ => Box::new(|| {}),
+        };
+
+        pool.install(|| op());
```
spacejam: one issue with doing this is that our total operating time is bound by how long it takes for one thread to generate all of the operations, which are then fed to the rayon queue. we may want to use raw threads that generate their operation before they perform their operation, which doesn't have to add up to the exact proportions that are configured as long as they are roughly equivalent, then do an atomic fetch_add on a shared counter, and if the result is above the configured number of total operations, return.

spacejam: this particular loop is probably faster than any individual operation in the tree, but I'm not sure if each iteration is like 32x faster, which is a common number of cores that are available to a server these days.

pmuens:
@spacejam thanks for the PR review 👍

I updated the code so that raw threads are used (according to our most-recent discussion).

spacejam:
looks good! this should be a lot faster now!

pmuens:
> looks good! this should be a lot faster now!

Great! Thanks the review and thanks for merging :+1:

PR #74, opened 2017-08-17
coco gc. closes #18. closes #19.

### LINKED ISSUE #18: add detached state to thread-local GC structure

### LINKED ISSUE #19: implement enroll - un-enroll - merge thread-local GC state logic

spacejam:


src/page/mod.rs
```diff
@@ -63,6 +63,27 @@ pub enum CacheEntry<M> {
 
 unsafe impl<M: Send> Send for CacheEntry<M> {}
 
+#[derive(Debug, Clone)]
+pub struct CasKey<P> {
+    ptr: *const stack::Node<CacheEntry<P>>,
+    tag: usize,
+}
```
spacejam: I've created this to act like a sort of "access key" that lets other threads later try to atomically update shared state with, but without having access to the pointer underneath. This lets us avoid using coco `Scope`'s in higher-level API's above the pagecache, while still having the advantages of epoch GC. I'll write more about this in a trip report for adapting to coco, since it's sort of a side-effect of the way scopes are handled there.

src/page/mod.rs
```diff
@@ -63,6 +63,27 @@ pub enum CacheEntry<M> {
...
+
+impl<'s, P> From<Ptr<'s, stack::Node<CacheEntry<P>>>> for CasKey<P> {
+    fn from(ptr: Ptr<'s, stack::Node<CacheEntry<P>>>) -> CasKey<P> {
+        CasKey {
+            ptr: ptr.as_raw(),
+            tag: ptr.tag(),
+        }
+    }
+}
```
spacejam: this lets coco shared atomic `Ptr`'s be converted to a `CasKey` as it crosses the boundary from coco-scoped land to the external users of the pagecache API (which should not be able to dereference any shared pointers)

src/page/mod.rs
```diff
@@ -63,6 +63,27 @@ pub enum CacheEntry<M> {
...
+
+impl<'s, P> Into<Ptr<'s, stack::Node<CacheEntry<P>>>> for CasKey<P> {
+    fn into(self) -> Ptr<'s, stack::Node<CacheEntry<P>>> {
+        unsafe { Ptr::from_raw(self.ptr).with_tag(self.tag) }
+    }
+}
```
spacejam: this lets us convert `CasKey`'s to coco `Ptr`'s so that we can pass the pointer out of the coco-scope protected code, and be used later for attempting to compare-and-swap a coco-protected variable.

src/page/page_cache.rs
```diff
@@ -99,13 +100,12 @@ impl<PM, P, R> PageCache<PM, LockFreeLog, P, R>
 
     /// Create a new page, trying to reuse old freed pages if possible
     /// to maximize underlying `Radix` pointer density.
-    pub fn allocate(&self) -> (PageID, *const stack::Node<CacheEntry<P>>) {
+    pub fn allocate(&self) -> (PageID, CasKey<P>) {
```
spacejam: all public-facing api's have no idea about coco scopes, or the details of the pointers that will later be used for CAS operations. the details are hidden away in the `CasKey` that they do have access to.

PR #76, opened 2017-08-20
sanitizers, reducing dependencies, conditional compiliation of deps, fix leaks. closes #59.

### LINKED ISSUE #59: add failing memory leak test script

spacejam:


spacejam:
@pmuens this PR adds the conditional dependencies to `Cargo.toml` for bench as well!

The stress tool is basically a specially tuned tool for sussing out concurrency and memory issues, by being configured to maximize overhead within the high-logic areas (small node sizes so they are split often, small cache threshold so things get dropped frequently, small consolidation threshold so pages get consolidated often, etc...). It is not about measuring latency histograms on realistic workloads, which is what the bench tool is for. Basically I needed something like this to use with the sanitizer tests. There is a bit of overlap, and feel free to just copy any parts of it that are useful, but it doesn't cover latency histograms, which is the main purpose of the bench tool.

pmuens:
> @pmuens this PR adds the conditional dependencies to Cargo.toml for bench as well!

Whoops. Looks like I missed that when skimming through the PR. Thanks for adding!

> The stress tool is basically a specially tuned tool for sussing out concurrency and memory issues, by being configured to maximize overhead within the high-logic areas (small node sizes so they are split often, small cache threshold so things get dropped frequently, small consolidation threshold so pages get consolidated often, etc...). It is not about measuring latency histograms on realistic workloads, which is what the bench tool is for. Basically I needed something like this to use with the sanitizer tests. There is a bit of overlap, and feel free to just copy any parts of it that are useful, but it doesn't cover latency histograms, which is the main purpose of the bench tool.

Gotcha! That makes sense 👍. Great addition!

spacejam:
I've spent the last couple days chasing down the remaining memory leaks. I also implemented safe reuse of PageID's by implementing a `PidDropper` object that the epoch system drops when it's safe, and in the `Drop` code it pushes the ID to the free stack for reuse. Things are looking a lot more reliable now, and I'm getting excited for telling a few more people about the system soon!

pmuens:
> I've spent the last couple days chasing down the remaining memory leaks. I also implemented safe reuse of PageID's by implementing a PidDropper object that the epoch system drops when it's safe, and in the Drop code it pushes the ID to the free stack for reuse. Things are looking a lot more reliable now, and I'm getting excited for telling a few more people about the system soon!

Great news! 👍

PR #80, opened 2017-08-22
Add bench statistics

### LINKED ISSUE #57: collect statistics on each operation in the benchmark tool
when executing operations, we want to know about:
1. throughput (requests / s)
1. latency (time per request)

For latency, we care about the latency distribution, which means we need to collect percentiles. This is actually a pretty interesting problem, since almost all existing solutions REALLY suck and use probabilistic approches that destroy accuracy. One approach that I like is:
* create a `Vec` of `AtomicUsize` that is pretty big. 
* create a funciton that maps from a measured value to a "slot number"
* each slot number corresponds to values that are ~1% larger than the ones in the previous one
* when measuring something, like latency, convert the latency to the slot number using the function, then look up the `AtomicUsize` and `fetch_add` 1 to it.
* when done with all measurements, you can scan over the counts to determine any percentile. it's good to know, at a minimum, 0th (fastest), 50th (median), 75th, 90th, 95th, 97.5th, 99th, 99.9th, 99.99th, 99.999th, and 100th (max). 

here's an example of a function that converts from a number to a slot number, and back: https://github.com/spacejam/loghisto/blob/master/metrics.go#L316

For throughput, just atomically increment a counter for each of the ops, and have a thread spit out the values every second, resetting it to 0. The latency percentiles can be for the duration of the entire benchmark, for now.

pmuens:
Closes #57

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 
 [features]
 default = []
-bench = ["clap", "num_cpus", "log", "rand", "rayon", "zstd", "env_logger"]
+bench = ["clap", "num_cpus", "log", "rand", "chrono", "rayon", "zstd", "env_logger"]
 stress = ["docopt", "chan-signal", "rayon", "rand", "zstd"]
 
 [[bin]]
@@ -49,6 +49,7 @@ clap = {version = "2.26", optional = true}
 rand = {version = "0.3", optional = true}
 docopt = {version = "0.8", optional = true}
 chan-signal = {version = "0.3", optional = true}
+chrono = {version = "0.4", optional = true}
 
 [dev-dependencies]
 quickcheck = "0.2"
--- a/src/bin/bench.rs
+++ b/src/bin/bench.rs
@@ -4,18 +4,21 @@ extern crate num_cpus;
 extern crate log;
 extern crate rand;
 extern crate sled;
+extern crate historian;
+extern crate chrono;
 
 use std::error::Error;
 use std::io::prelude::*;
 use std::process;
 use std::collections::HashMap;
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 use clap::{App, Arg};
 use rand::Rng;
+use historian::Histo;
 
 use sled::Tree;
 
@@ -189,6 +192,7 @@ fn perform_tree_operations(tree: Tree, config: Config) {
     let ops_per_thread = config.num_operations / config.num_threads; // TODO update to handle division errors
     let mut threads = Vec::new();
     let ops_per_second = AtomicUsize::new(0);
+    let histo = Arc::new(Histo::default());
 
     let tree = Arc::new(tree);
     let ops = Arc::new(ops);
@@ -210,8 +214,10 @@ fn perform_tree_operations(tree: Tree, config: Config) {
         let tree = tree.clone();
         let ops = ops.clone();
         let ops_per_second = ops_per_second.clone();
+        let histo = histo.clone();
         let t = thread::spawn(move || for _ in 0..ops_per_thread {
             let op_to_perform = get_operation_choice(&ops, sum_ops);
+            let now = Instant::now();
 
             match op_to_perform {
                 Some(&Op::Set) => perform_set_operation(&tree),
@@ -222,6 +228,13 @@ fn perform_tree_operations(tree: Tree, config: Config) {
                 _ => (),
             };
 
+            let us = chrono::Duration::from_std(now.elapsed())
+                .unwrap()
+                .num_microseconds()
+                .unwrap() as f64;
+
+            histo.measure(us);
+
             ops_per_second.fetch_add(1, Ordering::Relaxed);
         });
         threads.push(t);
@@ -230,6 +243,19 @@ fn perform_tree_operations(tree: Tree, config: Config) {
     for t in threads.into_iter() {
         t.join();
     }
+
+    println!("");
+    println!("0th: {}us", histo.percentile(0.).round() as usize);
+    println!("50th: {}us", histo.percentile(50.).round() as usize);
+    println!("75th: {}us", histo.percentile(75.).round() as usize);
+    println!("90th: {}us", histo.percentile(90.).round() as usize);
+    println!("95th: {}us", histo.percentile(95.).round() as usize);
+    println!("97.5th: {}us", histo.percentile(97.5).round() as usize);
+    println!("99th: {}us", histo.percentile(99.).round() as usize);
+    println!("99.9th: {}us", histo.percentile(99.9).round() as usize);
+    println!("99.99th: {}us", histo.percentile(99.99).round() as usize);
+    println!("99.999th: {}us", histo.percentile(99.999).round() as usize);
+    println!("100th: {}us", histo.percentile(100.).round() as usize);
 }
 
 fn perform_set_operation(tree: &Tree) {
--- a/src/bin/bench.rs
+++ b/src/bin/bench.rs
@@ -10,7 +10,9 @@ use std::io::prelude::*;
 use std::process;
 use std::collections::HashMap;
 use std::thread;
+use std::time::Duration;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 use clap::{App, Arg};
 use rand::Rng;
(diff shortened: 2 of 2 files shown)


src/bin/bench.rs
```diff
@@ -185,19 +188,40 @@ fn perform_tree_operations(tree: Tree, config: Config) {
...
     for i in 0..config.num_threads {
         let tree = tree.clone();
         let ops = ops.clone();
+        let ops_per_second = ops_per_second.clone();
+        let tx = tx.clone();
         let t = thread::spawn(move || for _ in 0..ops_per_thread {
             let op_to_perform = get_operation_choice(&ops, sum_ops);
 
+            ops_per_second.fetch_add(1, Ordering::SeqCst);
+            tx.send(ops_per_second.load(Ordering::SeqCst));
```
spacejam: for any of these counters, we can use the `Relaxed` memory ordering, which is still atomic, but does not slow down other threads so much.

spacejam: no need to use a channel, since that also forces the receiving thread to be a bottleneck for everyone.

src/bin/bench.rs
```diff
@@ -185,19 +188,40 @@ fn perform_tree_operations(tree: Tree, config: Config) {
...
     let ops_per_thread = config.num_operations / config.num_threads;
     let mut threads = Vec::new();
+    let statistics = Statistics::new();
+    let (tx, rx) = channel();
+
+    let statistics = Arc::new(statistics);
+    thread::spawn(move || {
+        loop {
+            // TODO how do I get the most recent value here rather than receiving one at a time?
+            let ops_per_second = rx.recv().unwrap();
```
spacejam: rather than using a channel, you can just have other threads do `fetch_add` on the counter to increment it, and then this thread just calls `swap(0)` on the counter once per second, which resets the value to 0 and returns the previous count.

src/bin/bench.rs
pmuens: Thanks! 🤔 yes, that makes 👍

src/bin/bench.rs
pmuens: Gotcha! Really elegant solution!

Completely forgot that I can share the `ops_per_second` in more than one thread.

pmuens:
Ok. Thanks for the review @spacejam 👍

That was really helpful! I was able to resolve the PR comments. Will push an updated, way more performant version soon!

pmuens:
FYI: @spacejam just rebased and pushed a couple of updates for the `throughput` statistic.

src/bin/bench.rs
```diff
@@ -185,19 +187,40 @@ fn perform_tree_operations(tree: Tree, config: Config) {
...
+
+    // thread which spits out bench-related results every 1 second
+    {
+        let ops_per_second = ops_per_second.clone();
+        let ops_per_second_vec = ops_per_second_vec.clone();
+        thread::spawn(move || {
+            loop {
+                thread::sleep(Duration::from_secs(1));
+                let ops_per_second = ops_per_second.swap(0, Ordering::Relaxed);
+                let mut ops_per_second_vec = ops_per_second_vec.lock().unwrap();
```
spacejam: I think we can probably just print out the throughput each second as it happens, rather than buffering it

pmuens: Good idea. This makes the code easier to write for me 😄!

So we get rid of the array to store all the throughputs and won't have an overall throughput after the benchmark is done?

spacejam:
@pmuens I've writte a simple library for recording histogram values and later printing percentiles:
https://docs.rs/historian/1.0.0/historian/struct.Histo.html

pmuens:
> @pmuens I've writte a simple library for recording histogram values and later printing percentiles:
https://docs.rs/historian/1.0.1/historian/struct.Histo.html

Thanks for that @spacejam 👍! Will update the PR so that `historian` is used. Outstanding naming BTW 😄

pmuens:
@spacejam so I just used the `historian` lib (which looks pretty awesome BTW 👍).

My current approach is to measure the elapsed nanos between the start and the end of the `Tree` operation (you can see the first attempt in the last commit).

However it looks like the `historian::radix::Node` cannot be safely shared between threads:

```bash
root@405c91c6f034:/app# cargo build --bin bench --features="bench" && target/debug/bench --num-operations 10000 --set 10 --get 20
   Compiling rsdb v0.10.0 (file:///app)
error[E0277]: the trait bound `*mut std::sync::atomic::AtomicUsize: std::marker::Send` is not satisfied in `historian::radix::Node`
   --> src/bin/bench.rs:218:17
    |
218 |         let t = thread::spawn(move || for _ in 0..ops_per_thread {
    |                 ^^^^^^^^^^^^^ `*mut std::sync::atomic::AtomicUsize` cannot be sent between threads safely
    |
    = help: within `historian::radix::Node`, the trait `std::marker::Send` is not implemented for `*mut std::sync::atomic::AtomicUsize`
    = note: required because it appears within the type `historian::radix::Node`
    = note: required because of the requirements on the impl of `std::marker::Send` for `coco::epoch::atomic::Atomic<historian::radix::Node>`
    = note: required because it appears within the type `historian::radix::Radix`
    = note: required because it appears within the type `historian::Histo`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<historian::Histo>`
    = note: required because it appears within the type `[closure@src/bin/bench.rs:218:31: 236:10 ops_per_thread:usize, ops:std::sync::Arc<std::vec::Vec<(Op, usize)>>, sum_ops:usize, tree:std::sync::Arc<rsdb::Tree>, histo:std::sync::Arc<historian::Histo>, ops_per_second:std::sync::Arc<std::sync::atomic::AtomicUsize>]`
    = note: required by `std::thread::spawn`
```

Is my attempt correct? Does the `historian` lib need some updates so that it can be safely shared between threads?

spacejam:
@pmuens whoops, added the right markers for using it across threads.

Cargo.toml
```diff
@@ -42,6 +42,7 @@ clap = {version = "2.26", optional = true}
 rand = {version = "0.3", optional = true}
 docopt = {version = "0.8", optional = true}
 chan-signal = {version = "0.3", optional = true}
+historian = {version = "1.0.1", optional = true}
```
spacejam: bump this to `2.0` (leave off the last part for it to track the latest patch)

src/bin/bench.rs
```diff
@@ -184,18 +188,36 @@ fn perform_tree_operations(tree: Tree, config: Config) {
         (Op::Delete, config.delete),
         (Op::Cas, config.cas),
     ];
-
-    // TODO update to handle division errors
-    let ops_per_thread = config.num_operations / config.num_threads;
+    let ops_per_thread = config.num_operations / config.num_threads; // TODO update to handle division errors
     let mut threads = Vec::new();
+    let ops_per_second = AtomicUsize::new(0);
+    let histo = Histo::default();
```
spacejam: generally add the `Arc` inline, rather than redefining the variable below

pmuens:
@spacejam thanks for the review. Just pushed some requested in the review.

The benchmarking tool now prints all the percentiles (as requested in #57) at the end of the execution:

```bash
running benchmarking suite...
throughput: 1860
throughput: 1795
throughput: 1714
throughput: 1589
throughput: 1407
throughput: 1350

0th: 103776ns
50th: 994504ns
75: 1369559ns
90th: 2043142ns
95th: 2730512ns
97.5th: 3368573ns
99th: 5230408ns
99.9th: 7725212ns
99.99th: 39428520ns
99.999th: 51136034ns
100th: 51136034ns
```

Not sure if `ns` is too precise. I converted it to `ms`, but IMHO that's to imprecise.

spacejam:
microseconds (us) is probably the appropriate granularity for the time being. 14ms is crazy, I'll bet that's when operations coincide with a snapshot

pmuens:
> microseconds (us) is probably the appropriate granularity for the time being. 14ms is crazy, I'll bet that's when operations coincide with a snapshot

Thanks for the feedback @spacejam 👍. I'll update the PR accordingly.

**Aside:** One thing to note about the benchmarking results is that I run this inside of Docker 😬 (and I haven't allocated too much resources for the containers)

pmuens:
FYI: @spacejam just updated the PR so that microseconds are used.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ readme = "README.md"
 
 [features]
 default = []
-bench = ["clap", "num_cpus", "log", "rand", "rayon", "zstd", "env_logger"]
+bench = ["clap", "num_cpus", "log", "rand", "historian", "time", "rayon", "zstd", "env_logger"]
```
spacejam: since I added historian as a dep on the main crate, this doesn't need to be in the feature array

Cargo.toml
```diff
@@ -45,6 +45,8 @@ clap = {version = "2.26", optional = true}
 rand = {version = "0.3", optional = true}
 docopt = {version = "0.8", optional = true}
 chan-signal = {version = "0.3", optional = true}
+historian = {version = "2.0", optional = true}
```
spacejam: can nuke this now since historian is a dependency of the crate always

src/bin/bench.rs
```diff
@@ -184,18 +190,35 @@ fn perform_tree_operations(tree: Tree, config: Config) {
...
     let tree = Arc::new(tree);
     let ops = Arc::new(ops);
+    let ops_per_second = Arc::new(ops_per_second);
+
+    // thread which spits out bench-related results every 1 second
+    {
+        let ops_per_second = ops_per_second.clone();
+        thread::spawn(move || {
+            loop {
+                thread::sleep(Duration::seconds(1).to_std().unwrap());
```
spacejam: `thread::sleep(Duration::from_secs(1))`

spacejam:
Looking good! As soon as the minor things I commented on are in this is good to go!

src/bin/bench.rs
pmuens: Thanks for commenting 👍!

I had that before, but now since I'm using the `Duration` struct from `time` crate I had a naming conflict. I decided to pick one to keep the usage consistent and it this case it was the `Duration` struct from `time`.

pmuens: Lol. Looks like `time` is no longer maintained (https://github.com/rust-lang-deprecated/time#notes). Maybe I switch to `chronos`. I wish that `cargo` shows the READMEs...

pmuens:
Thanks for the review @spacejam 👍

I'll look through the comments, fix them and rebase the PR so that it's ready to be merged.

pmuens:
@spacejam just updated and rebased according to your PR review.

Here's how the output looks like:

```bash
root@0db01d72d732:/app# cargo build --bin bench --features="bench" && target/debug/bench --num-operations 10000 --set 10 --get 20
    Finished dev [unoptimized + debuginfo] target(s) in 0.4 secs
running benchmarking suite...
throughput: 1675
throughput: 1058
throughput: 854
throughput: 1154
throughput: 1860
throughput: 1344
throughput: 501

0th: 51us
50th: 339us
75th: 705us
90th: 1379us
95th: 6062us
97.5th: 9604us
99th: 13766us
99.9th: 388480us
99.99th: 568069us
99.999th: 568069us
100th: 568069us
throughput: 1554
```

PR #84, opened 2017-08-23
Add padding for keys and values

pmuens:
Adds padding for keys and values so that it's more likely to e.g. `get` a previously inserted tree entry.

This way we don't have to seed the tree with dummy data before running the benchmarking tool.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ readme = "README.md"
 
 [features]
 default = ["rayon", "zstd", "libc"]
-bench = ["clap", "num_cpus", "env_logger", "rayon", "zstd"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -11,7 +11,7 @@ readme = "README.md"
 
 [features]
 default = ["rayon", "zstd"]
-bench = ["clap", "num_cpus", "env_logger", "rayon", "zstd"]
```
spacejam: should keep zstd for realistic tests with compression

pmuens: Ok. I thought that this list should just include all the direct dependencies of the `bench` tool. 

I just amend the last commit with the requested changed.

pmuens:
Resolved all the conflicts

PR #92, opened 2017-08-25
simplify pagecache semantics. closes #91.

### LINKED ISSUE #91: investigate merge operator-like semantics for pagecache
try reimplementing the pagecache with a single "type" and use set/merge instead of prepend/replace. this allows us to more effectively cache materialized views without duplicating stored data, and allowing us to return zero-copy references if the user provides a scope, but we need to think about that later.

spacejam:


spacejam:
This will dramatically improve our performance, once we cache calls to merge for every get request, as we no longer need to know about this other MaterializedPage.

PR #93, opened 2017-08-27
major pagecache overhaul, quickcheck, lots of bugs fixed. closes #77. closes #49. 

### LINKED ISSUE #49: quickcheck on pagecache

### LINKED ISSUE #77: tune materializer calls
calls to the page materializer take a lot of time, try to cache results from it so we don't spend as much cpu calling it on the same hot pages over and over.

spacejam:


src/config.rs
```diff
@@ -31,6 +31,7 @@ pub struct Config {
     flush_every_ms: Option<u64>,
     snapshot_after_ops: usize,
     snapshot_path: Option<String>,
+    cache_fixup_threshold: usize,
```
spacejam: this is the number of unmerged partial pages in cache that can exist before an attempt to squish them together into a cached `MergedResident`

src/log/iobuf.rs
```diff
@@ -258,14 +258,15 @@ impl IoBufs {
...
+            let reservation_offset = log_offset + buf_offset as LogID;
 
             return Reservation {
                 idx: idx,
                 iobufs: self,
                 data: buf,
                 destination: destination,
                 flushed: false,
-                base_disk_offset: log_offset as LogID,
+                reservation_offset: reservation_offset,
```
spacejam: this was a stupid bug that went undetected because the pagecache wasn't well tested before, where the log id for a reservation was just the ID of the base disk offset. it's amazing that the tree was working with recovery tests...

src/page/mod.rs
```diff
@@ -43,6 +43,9 @@ pub trait Materializer {
 /// Points to either a memory location or a disk location to page-in data from.
 #[derive(Debug, Clone, PartialEq)]
 pub enum CacheEntry<M: Send + Sync> {
+    /// A cache item that contains the most recent fully-merged page state, also in secondary
+    /// storage.
+    MergedResident(M, LogID),
```
spacejam: this allows us to do less cpu-intensive merging when retrieving pages, at the cost of higher contention for mutations

spacejam: in the future it may be possible to do this without effecting contention by making pages mergeable without effecting the stack pointers

src/page/page_cache.rs
```diff
@@ -316,36 +323,103 @@ impl<PM, P, R> PageCache<PM, LockFreeLog, P, R>
...
+        let to_evict = self.lru.accessed(pid, size);
+        self.page_out(to_evict, scope);
+
+        if lids.len() > self.config().get_page_consolidation_threshold() {
+            match self.set(pid, head.into(), merged.clone()) {
+                Ok(new_head) => head = new_head.into(),
+                Err(None) => return None,
+                _ => {}
+            }
+        } else if fix_up_length >= self.config().get_cache_fixup_threshold() {
```
spacejam: this is the logic for swapping out a stack for a merged cache stack

src/log/iobuf.rs
pmuens: 😃 Nice catch 👍!

tests/test_pagecache.rs
```diff
@@ -64,3 +107,319 @@ fn basic_recovery() {
...
+fn quickcheck_pagecache_works() {
+    QuickCheck::new()
+        .gen(StdGen::new(rand::thread_rng(), 1))
+        .tests(50)
+        .max_tests(100)
+        .quickcheck(prop_pagecache_works as fn(OpVec, u8) -> bool);
+}
+
+#[test]
+fn test_pagecache_bug_1() {
```
pmuens: Love that regression-test schema (with postmortem!) 💯

Cargo.toml
```diff
@@ -12,7 +12,7 @@ readme = "README.md"
 [features]
 default = ["rayon", "zstd", "libc"]
 bench = ["clap", "num_cpus", "env_logger", "rayon", "zstd"]
-race = ["docopt", "chan-signal", "rand", "rayon", "env_logger"]
+race = ["docopt", "chan-signal", "rand", "rayon"]
```
pmuens: Just a quick question about the `features` definition here.

Should all the dependencies (`extern crate`s) for the binary be listed here?

The [documentation](http://doc.crates.io/manifest.html#the-features-section) is a little bit vague about that.

Why should I list optional dependencies here? For me only "hard" dependencies make sense 🤔

PR #101, opened 2017-08-31
a bunch of tuning and some reorganizing. closes #78.

### LINKED ISSUE #78: tune Lru
calls to the Lru bookkeeping structures are inefficient. should use simpler data structures.

spacejam:
we're now up to over 200k op/s on a crappy laptop, hoping to 10x it over time

this is the first strike in the clone wars https://github.com/spacejam/rsdb/issues/100


src/dll.rs
```diff
@@ -0,0 +1,195 @@
+/// A simple doubly-linked list for use in the `Lru`
```
spacejam: this is the main thing that powers an `Lru` now, protected by a mutex and heavily sharded, so that the threads are using like 256 of these for now to avoid contention.

src/ds/art.rs
```diff
@@ -0,0 +1,79 @@
+/// An adaptive radix tree.
```
spacejam: this is an experiment in reducing pointer density in the radix tree underlying many of our structures, not currently in use anywhere yet, barely started, mostly incomplete!

src/ds/dll.rs
```diff
@@ -0,0 +1,196 @@
+/// A simple doubly-linked list for use in the `Lru`
```
spacejam: this is the main structure powering the `Lru` now, a simple doubly linked list, sharded 256x and protected by a `Mutex`, so contention isn't too high.

src/hash/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
+
 mod hash;
```
spacejam: no longer using hashing for the Lru, so hash is currently unused, but I suspect it will find a use later on

src/tree/frag.rs
```diff
@@ -26,6 +26,13 @@ impl Frag {
             _ => None,
         }
     }
+
+    pub fn into_base(self) -> Option<(tree::Node, bool)> {
+        match self {
+            Frag::Base(base, root) => Some((base, root)),
+            _ => None,
+        }
+    }
```
spacejam: this lets us avoid cloning the frag to pull out the base, if we're able to have ownership over the frag

src/tree/mod.rs
```diff
@@ -492,9 +493,9 @@ impl Tree {
                 // println!("after: {:?}", self);
             }
 
-            path.push((node.clone(), cas_key));
+            path.push((node, cas_key));
 
-            match node.data {
+            match path.last().unwrap().0.data {
```
spacejam: don't clone the node, instead just put it in the returned vec and reference it

spacejam:
hmmmm... seem to have caused a tsan false positive by using raw pointers the `Dll`! looking into the possibility of adding `__attribute__((no_sanitize("thread")))` from LLVM as a supported attribute in rust... otherwise we may be better off biting the bullet and making the pointer operations atomic just to satisfy tsan, since we get a ton of utility with it elsewhere here...

spacejam:
Actually it wasn't a false positive! I was failing to null the prev pointer on a `Node` when unlinking it, which caused some nodes to be freed multiple times. However, I did find a way to suppress certain checks if we need to do so in the future: https://github.com/google/sanitizers/wiki/ThreadSanitizerSuppressions

pmuens:
> Actually it wasn't a false positive! I was failing to null the prev pointer on a Node when unlinking it, which caused some nodes to be freed multiple times. However, I did find a way to suppress certain checks if we need to do so in the future: https://github.com/google/sanitizers/wiki/ThreadSanitizerSuppressions

Nice find! Good to hear that it wasn't a false positive 👍. Thanks for sharing the link!

PR #104, opened 2017-09-03
tune page_in a bit

spacejam:
1.5m failed reads/s on an empty tree
750k reads/s 4 threads all doing get's with 8-byte keys and 0 byte values
230k ops/s with a workload that does 1 type of op per thread
crappy laptop, 2 cores with hyperthreading
probably enough tuning for the short term, time to focus more on release engineering now

### DIFF
--- a/src/page/page_cache.rs
+++ b/src/page/page_cache.rs
@@ -386,6 +386,12 @@ impl<PM, P, R> PageCache<PM, P, R>
             .rev()
             .collect();
 
+        // Short circuit merging and fix-up if we only
+        // have one frag.
+        if combined.len() == 1 {
+            return Some((combined[0].clone(), head.into()));
+        }
+
         let merged = self.t.merge(&*combined);
 
         let size = std::mem::size_of_val(&merged);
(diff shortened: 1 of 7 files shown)


src/page/page_cache.rs
```diff
@@ -386,6 +386,12 @@ impl<PM, P, R> PageCache<PM, P, R>
             .rev()
             .collect();
 
+        // Short circuit merging and fix-up if we only
+        // have one frag.
+        if combined.len() == 1 {
+            return Some((combined[0].clone(), head.into()));
+        }
```
spacejam: this is the meat of this particular optimization

src/tree/mod.rs
```diff
@@ -16,8 +16,6 @@ pub use self::frag::*;
 pub use self::data::*;
 pub use self::node::*;
 
-pub const FANOUT: usize = 2;
```
spacejam: this was a huge facepalm, we were basically hardcoded to a binary tree lol

pmuens: Lol! That's a good one! 😄

PR #106, opened 2017-09-03
two bugfixes in page_in

spacejam:
* fix bug in short-circuit read logic
* fix bug in page_in logic for storing retrieved fragments

### DIFF
--- a/src/page/page_cache.rs
+++ b/src/page/page_cache.rs
@@ -360,6 +360,12 @@ impl<PM, P, R> PageCache<PM, P, R>
             return None;
         }
 
+        // Short circuit merging and fix-up if we only
+        // have one frag.
+        if merged_resident && to_merge.len() == 1 {
+            return Some((to_merge[0].clone(), head.into()));
+        }
+
         let mut fetched = Vec::with_capacity(lids.len());
 
         // Did not find a previously merged value in memory,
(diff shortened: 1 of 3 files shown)


src/page/page_cache.rs
```diff
@@ -386,12 +392,6 @@ impl<PM, P, R> PageCache<PM, P, R>
             .rev()
             .collect();
 
-        // Short circuit merging and fix-up if we only
-        // have one frag.
-        if combined.len() == 1 {
-            return Some((combined[0].clone(), head.into()));
-        }
-
```
spacejam: this could return a fragment that was paged-out and just read off the log before we have a chance to store it

src/page/page_cache.rs
```diff
@@ -404,7 +404,9 @@ impl<PM, P, R> PageCache<PM, P, R>
                 Err(None) => return None,
                 _ => {}
             }
-        } else if fix_up_length >= self.config().get_cache_fixup_threshold() {
+        } else if !fetched.is_empty() ||
+                   fix_up_length >= self.config().get_cache_fixup_threshold()
```
spacejam: we were not actually remembering page fragments that we retrieved from disk

src/page/page_cache.rs
```diff
@@ -750,6 +752,7 @@ impl<PM, P, R> PageCache<PM, P, R>
                 // after the scope passes, these lids will
                 // have their disk space hole punched.
                 scope.defer_drop(dropper.into_ptr(scope));
+                scope.flush();
```
spacejam: because the act of punching a hole in a log could be expensive, it's better to trigger a flush here to spread out the work over time, rather than stick one thread with a ton of holes to punch in the log

PR #119, opened 2017-09-06
add feature flag for using O_DIRECT on writes. closes #111.

### LINKED ISSUE #111: investigate O_DIRECT for writing buffers to disk
we only write serialized data to disk, which will stay in memory as long as the PageCache thinks it's useful. maybe we should bypass the OS's pagecache on writes.

spacejam:
O_DIRECT skips the OS pagecache entirely, basically doing DMA. It makes it so the OS does not cache any data in memory about files, as it normally does. This is sometimes useful if the data on disk is different from what the application actually uses (compression, serialization, etc...) and there's a userspace cache already (like what we have). A downside is you can get WORSE performance if you mix O_DIRECT and non-O_DIRECT work, and you have to align all operations to the device's specified alignment (usually 512, but sometimes not!)

This work basically gives us a quick switch to flip when we want to give it a spin!

Note that it's not correct yet, since the 512 hardcoded alignment could be different sometimes, so we need to check that dynamically in the future.

### DIFF
--- a/src/log/iobuf.rs
+++ b/src/log/iobuf.rs
@@ -1,4 +1,5 @@
 use std::io::{Seek, Write};
+use std::ptr;
 
 #[cfg(feature = "zstd")]
 use zstd::block::compress;
@@ -126,8 +127,12 @@ impl IoBufs {
         options.create(true);
         options.write(true);
 
-        //use std::os::unix::fs::OpenOptionsExt;
-        //options.custom_flags(libc::O_DIRECT);
+        #[cfg(target_os = "linux")]
+        #[cfg(feature = "o_direct_writer")]
+        {
+            use std::os::unix::fs::OpenOptionsExt;
+            options.custom_flags(libc::O_DIRECT);
+        }
 
         let file = options.open(path).unwrap();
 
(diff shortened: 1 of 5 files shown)


src/log/iobuf.rs
```diff
@@ -426,13 +441,32 @@ impl IoBufs {
         );
 
         let res_len = offset(header) as usize;
+
+        #[cfg(feature = "o_direct_writer")]
+        let pad_len = {
+            let remainder = res_len % 512;
+            if remainder == 0 { 0 } else { 512 - remainder }
+        };
```
spacejam: here we have 2 choices: add a bunch of padding in-between log entries, or do extra buffering so we can keep things extra squished. Adding some padding is faster, and still correct since our log iterator will just skip over zeroed out entries anyway.

spacejam: (must do everything on the 512 (or some other) byte boundaries with O_DIRECT!)

spacejam:
@pmuens many! it's totally problematic in a wide variety of situations that we will want to turn it off for. BUT if we get things properly tuned, it should be a huge performance improvement in a reasonable subset of deployments.

pmuens:
> @pmuens many! it's totally problematic in a wide variety of situations that we will want to turn it off for. BUT if we get things properly tuned, it should be a huge performance improvement in a reasonable subset of deployments.

That makes sense. Sounds like a good tradeoff 👍

PR #120, opened 2017-09-07
extract c api into its own crate

spacejam:


spacejam:
this will keep clutter down, and make testing this on its own cleaner

PR #123, opened 2017-09-07
add rudimentary python binding

spacejam:


lib/python/alice_check.py
```diff
@@ -0,0 +1,19 @@
...
+dirlist = os.listdir('.')
+
+assert("ALICE.data" in dirlist)
+
+c = Conf()
+c.path(b"ALICE.data")
+
+t = c.tree()
+
+assert(t.get(b"k1") == b"v1")
```
spacejam: this check is run automatically by ALICE when testing for corruption bugs, accepting the test directory as the first argument, which the ALICE framework "corrupted" in some way, to test recovery

lib/python/alice_populate.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python3
+from sled import Conf
+
+c = Conf()
+c.path(b"ALICE.data")
+t = c.tree()
+t.set(b"k1", b"v1")
+t.close()
```
spacejam: this little script just writes a single key to the file `ALICE.data` which the ALICE framework will later permute and feed to the checker for verification

lib/python/rsdb.py
```diff
@@ -0,0 +1,109 @@
...
+sled.sled_get.restype = c_char_p
+
+sled.sled_scan.argtypes = (c_void_p, c_char_p, c_size_t, POINTER(c_size_t))
+sled.sled_scan.restype = c_void_p
+
+sled.sled_set.argtypes = (c_void_p, c_char_p, c_size_t, c_char_p, c_size_t)
+sled.sled_set.restype = None
+
+sled.sled_del.argtypes = (c_void_p, c_char_p, c_size_t)
+sled.sled_del.restype = None
```
spacejam: from python, we need to describe the type signatures in this way for `ctypes` to properly transform data before calling the c-compatible library. this will later probably use one of the other FFI libraries to work in a more flexible way, but for now it's cool to be calling this from python :)

lib/python/rsdb.py
```diff
@@ -0,0 +1,109 @@
...
+                          c_char_p, c_size_t,  # old
+                          c_char_p, c_size_t,  # new
+                          POINTER(c_char_p), POINTER(c_size_t),  # actual ret
+                          )
+sled.sled_cas.restype = c_ubyte
+
+
+class Conf:
+    def __init__(self):
+        self.ptr = c_void_p(sled.sled_create_config())
```
spacejam: @matthiasn here's how the config pointer is created

lib/python/rsdb.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def tree(self):
+        tree_ptr = sled.sled_open_tree(self.ptr)
+        return Tree(c_void_p(tree_ptr))
```
spacejam: @matthiasn this is what creates the `Tree` from the configuration so far

lib/python/rsdb.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def path(self, path):
+        sled.sled_config_set_path(self.ptr, path)
```
spacejam: @matthiasn this configures the path

lib/python/rsdb.py
```diff
@@ -0,0 +1,109 @@
...
+    def set(self, key, val):
+        sled.sled_set(self.ptr, key, len(key), val, len(val))
+
+    def get(self, key):
+        vallen = c_size_t(0)
+        ptr = sled.sled_get(self.ptr, key, len(key), byref(vallen))
+        return ptr[:vallen.value]
+
+    def delete(self, key):
+        sled.sled_del(self.ptr, key, len(key))
```
spacejam: @matthiasn set, get and del are probably the first 3 operations to support from the node module

PR #125, opened 2017-09-07
initial work for node binding

matthiasn:
work in progress


spacejam:
@matthiasn this might be an approach that could work for passing a pointer to node and back:
https://play.rust-lang.org/?gist=31afbb7d9cd897375ee6dfb72a647d3d&version=undefined

```
#[derive(Debug)]
struct Thing {
    val: usize,
}

fn main() {
    // make the thing
    let thing = Thing { val: 5 };
    println!("original thing: {:?}", thing);
    
    // throw it on the heap, make a raw pointer from it
    let thing_ptr = Box::into_raw(Box::new(thing));
    
    // cast the pointer to a usize, then into a string
    let thing_ptr_str = format!("{}", thing_ptr as usize);
    println!("thing_ptr_str: {}", thing_ptr_str);
    
    // convert the string back to a usize
    let ptr_from_str = thing_ptr_str.parse::<usize>().unwrap();
    println!("ptr_from_str: {}", ptr_from_str);
    
    // use the converted pointer
    let ptr = ptr_from_str as *mut Thing;
    unsafe {
        println!("dereferenced raw pointer: {:?}", *ptr);
    }
}
```

spacejam:
Merging this for now! Awesome work!

PR #131, opened 2017-09-09
tune write buffering. closes #105.

### LINKED ISSUE #105: asynchronously punch holes in aborted log reservations
log reservations that have been aborted cannot be fallocate'd away until their zeroes are on-disk.

spacejam:


spacejam:
this pushes responsibility for calling fallocate on the invalidated log entries to another thread. this doesn't really improve the performance issues with using fallocate, but it does let us defer clearing up space in failed flushes. next step is to actually fix the fallocate performance issues.

PR #143, opened 2017-09-20
major storage overhaul. closes #135. closes #141. closes #127.

### LINKED ISSUE #127: use cond var in make_stable rather than busywaiting

### LINKED ISSUE #135: log cleaning: for real this time
the biggest problem with this database right now is the fact that it grows unbounded on disk with operations. if you write a single key over and over, it will fill up the disk. the fallocate hack doesn't cut it, and we have to do a ton of work, as we issue calls to fallocate every time we replace a page, for each chunk on disk of updates. this can mean 50k calls for what was originally an 8mb flush.

the llama and bw tree papers refer to [the LFS paper](https://people.eecs.berkeley.edu/%7Ebrewer/cs262/LFS.pdf) as a way to work around this issue. they claim to be using a single circular buffer with a simple rewriting policy, but this seems like it would have really nasty write amplification properties. the LFS paper goes into detail about reducing write amplification, and talking about ways of keeping performance high. the microsoft papers claim that the overhead of adding the circular gc is in the single-digits, but maybe we can do better than this. unbounded, constant write amplification is $$$, particularly at high scales.

desired properties:
* over time, squishes old entries together so they are copied less and less often
* when writing a stable set of keys over and over, with constant value sizes, the on-disk size should have an upper bound

components:
* a priority queue which lets you remove and replace elements when priority changes
* an atomic LSN counter

implementation:
* write log segments at the configured power of 2 at a time, padding the end with zeroes
* give all updates an LSN
* give written log segments a reference to the previous written segment, and the next segment for traversing during recovery
* when we replace a page, mark each segment that it belongs to as being at lower utilization. this can be represented as a fraction of the initial total number of pages that were included.
* when the segment drops below a configurable threshold, it enters a priority queue of reuse candidates
* when we write a log segment, we need to know the next log segment to write as a pointer from the current one for traversing during recovery. so, we have a short known-offset queue for log writes. when we cannot pop the priority queue to populate the future write offset queue, we increment the tip of the log and add it to the end.
* when we pop a segment off the priority queue, if there are still live page fragments inside it, we need to rewrite compacted pages before using it. that means we always need to be thinking a little bit ahead. when we have fewer than a configurable number of completely free buffers, we can create them by defragmenting existing pages by copying them or by allocating a new segment at the end of the log.
* when we checkpoint periodically, we serialize the current priority queue at a specific log offset. we mark the last written segment to replay recovery from after reading in the checkpoint. the recovery traverses segments based on the next pointer. if, during recovery, the next pointer is past the end of the log, or the entries at the referred to location have LSN's that are lower than the entries in the previous segment, we have reached the end of the recovery.

### LINKED ISSUE #141: calls to flush should not waste space in segment-oriented write scheme
while implementing log cleaning, the io buffers fill entire segments at a time, which are then managed by the segment accountant. when other threads call flush, the entire segment gets rolled. this wastes up to 8mb on disk, filling the disk up crazy fast. this can be fixed by switching the lock-free log buffer header from being a current offset to a remainder, and then either grabbing a new segment offset if the segment is like >90% full, or providing an incremented base offset from within the last segment if that has 10% or more remaining.

spacejam:
This is a huge change to how storage is handled. Before, we always appended new entries to the log, and over time issues fallocate calls to do "hole punching" to tell the filesystem to deallocate stale data, keeping the real size of the file low even though the write offset keeps increasing. This didn't work because we write in 8mb blocks, but that turns into 50k random fallocate calls, and it's not compatible with osx or windows.

This uses ideas from the [LFS paper](https://people.eecs.berkeley.edu/%7Ebrewer/cs262/LFS.pdf) to split the storage file into "segments". Each segment keeps track of the pages that have live fragments inside. A page could have fragments scattered across several segments. A segment contains fragments for many pages. When a page is replaced and rewritten, the old locations for fragments are no longer necessary to keep around, so we use a `SegmentAccountant` to keep track of the usage of each segment over time. Once a segment reaches a particular threshold of emptiness, the `SegmentAccountant` will flag its pages as candidates for accelerated merging and rewriting to another segment. When a segment is completely empty, it is elligible to be rewritten with new data.

- [x] implement segment-based allocation and usage tracking
- [x] snapshot and restore the `SegmentAccountant`
- [x] safely delay segment reuse until all threads are guaranteed to be done with `LogID`'s that reference into it
- [x] safely delay segment freeing by always having more free segments than IO buffers, enforcing FIFO 
- [x] write backlinks into segments to detect missing previous chunks during recovery
- [x] write a trailer to each segment after fsync'ing it to detect torn segments
- [x] proactively rewrite pages based on `SegmentAccountant`'s priority list
- [x] burn-in with stress tests, hard thread jugglers, measure steady disk + memory + throughput
- [x] improve documentation
- [x] refactor


spacejam:
Persisting to segments can happen out-of-order, and with an arbitrary delay after we tell the `SegmentAccountant` we've tried to rewrite a page into an IO buffer. We need to address a couple of subtle correctness hazards before they bite anyone:

1. just because the `SegmentAccountant` knows that all pages resident to a segment have been put into IO buffers destined for other segments, we can't reuse the previous segment until we know the replacements are actually stable.
2. just because a segment's pages have been relocated to other segments and are stable, it doesn't mean there aren't threads that have references to the previous `LogID`'s for the page fragments in that segment.

I think the easiest way to guarantee safety is to do the following:
1. ensure that the queue of free segments is always longer than the configured number of IO buffers. Writes block on IO buffers being available, and they do not become available again until the previous write is written and fsync'd. By having more segments on-deck than IO buffers, we guarantee that segment reuse occurs only after all pending writes that will cause the segment to become free in the first place are definitely written and synced on disk. This is a fun juggling act!
2. thread dangling references are already solved by using the `coco` (and eventually the next generation `crossbeam-epoch` library) along with a "dropper" like what is being done for `PageID`'s already. The segment dropper will use the epoch-style GC to only add a segment to the free list after all threads that have accessed shared state (like `LogID`'s) are done with their work.

We need 2 AND 1 because there may be a long-running thread that is operating on shared state for a while, and could theoretically operate for several full cycles of the IO buffers.

src/log/iobuf.rs
```diff
@@ -617,3 +797,15 @@ fn bump_offset(v: u32, by: u32) -> u32 {
...
+fn debug_delay() {
+    #[cfg(test)]
+    {
+        use rand::{Rng, thread_rng};
+
+        if thread_rng().gen_weighted_bool(1000) {
+            std::thread::sleep(std::time::Duration::from_millis(100));
+        }
+    }
+}
```
spacejam: This thing (which gets fully elliminated from non-test code due to the `inline(always)` and `cfg(test)`) is a nice interleaving exerciser, ensuring we get great coverage for the lock-free algorithm in this system.

spacejam:
because segments could be written out-of-order, during recovery we need to be able to detect when a segment was missing or if a segment was torn, so we can halt recovery at the contiguous tip.

1. skipped segments can be solved by forcing writes to be sequential and non-concurrent, but this prevents multithreaded writes. the main problem is differentiating between replaced segments and missing segments. we could keep an ever-growing sequence of previous segments that each segment replaced, but maybe we don't need to know that much history to be correct. if we are forcing pages to go through the free queue that is at least as long as the number of IO buffers, then it should be enough just to record the previous LSN in a header for a segment. then we get multithreaded writes!
2. torn segments can be detected by writing a trailer to each segment AFTER the rest of the segment was fsync'd.

spacejam:
At this point, there is a little bit of refactoring (especially the recovery path) and burn-in left to do, but the basic structure of the segmented storage system is more-or-less as it will be. Version `0.15.0` will not be released until the refactoring has been complete, along with more burn-in and some post-trauma tuning. After tests go green this will be merged, to get this work into master.

Issue #98, opened 2017-08-29
read-only database configuration option


spacejam:
we should add the ability to open a database in read-only mode, so that multiple processes may share an underlying storage file. this involves adding the option to the `Config` struct, setting the default to false, adding a line for it in the get/set/builder macro lower down in the same file, and then disabling `set`, `cas`, `del` on the `Tree` when it's true, and disabling `reserve`, `write`, `punch_hole` on the `Log` when it's true, and `set`, `merge`, `free`, and `allocate` on the `PageCache`.

spacejam:
closed by https://github.com/spacejam/sled/issues/98

PR #150, opened 2017-10-22
Read-only mode

dotcypress:
Some initial progress on #98.


spacejam:
awesome! thanks for adding this :) only requested change: would you mind making `readonly` snake_case?

dotcypress:
@spacejam Done

spacejam:
badass, thank you! This change will be available in version `0.15.0` as soon as I release it! There are probably a few more days of stress testing before I feel comfortable releasing it, as there were some really significant architectural changes that need to be burned in more

Issue #158, opened 2017-12-12
nuke in-snapshot segment accounting


spacejam:
segment membership can be recomputed from the snapshot page map, so we should not bother with this redundant + bug prone parsing!

spacejam:
fixed in #169

PR #169, opened 2017-12-29
simplify snapshot state

spacejam:
remove segment state from the `Snapshot` itself, recompute it on recovery from the page table. reduce bug space and concentrate responsibility.


spacejam:
closed https://github.com/spacejam/sled/issues/158

PR #192, opened 2018-01-11
verify that configuration is unchanged between openings. closes #134

### LINKED ISSUE #134: prepend storage file with metadata header
we should have a framed protobuf as the header for the main storage file, which contains information about the version of the system and what sort of compression was used, and in the future it will also contain stuff like merge operators or namespaces.

spacejam:


spacejam:
later this will include a whitelist that includes things that are safe to change

PR #215, opened 2018-01-31
Use HTTPS for merit badges, and GitHub raw link for embedded images

lambda:
The READMEs for sled and pagecache were rendering incorrectly on crates.io, due to some embedded merit badges with non-HTTPS links, and an image that wasn't actually linked to the raw image. Fix the links appropriately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # sled likes eating data! it's pre-alpha
 
 <p>
-  <img src="https://github.com/spacejam/rsdb/blob/master/art/tree_face.png" width="20%" height="auto" />
+  <img src="https://raw.githubusercontent.com/spacejam/sled/master/art/tree_face.png" width="20%" height="auto" />
 </p>
 
 [![Build Status](https://travis-ci.org/spacejam/sled.svg?branch=master)](https://travis-ci.org/spacejam/sled)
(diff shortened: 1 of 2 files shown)


spacejam:
Ahh I was wondering why this was happening, thank you very much, @lambda!

The tsan failure MIGHT be a false positive, since after refactoring pagecache out of sled, it seems to sometimes be triggered or not based on the way some memory gets laid out. Even adding a comment can trigger it. I'm going to drill a little more before merging, just to try to keep master green.

spacejam:
Yeah, after digging in more I'm fairly sure this is because TSAN does not pick up on raw memory fences, which trips up crossbeam_epoch [here](https://github.com/crossbeam-rs/crossbeam-epoch/blob/20701a5fc07db09a5eb0974a9d73c396e205da37/src/internal.rs#L116).

In the past, this was something I had to add a suppressions file for due to lazy_static relying on adjacent atomic operations for making non-atomic accesses on global static state safe. It's sort of weird that TSAN doesn't always go crazy on this for other concurrency tests that run though... I'm making a note to stay suspicious of this until I can explain why it only fails with particular memory arrangements.

So, I've added a suppressions file that I believe will silence this false positive: [commit](https://github.com/spacejam/sled/commit/425f4ba64bc3d7c67877ef800bec6777c9fddf44).

Thanks again for fixing this! It was a real wart!

PR #217, opened 2018-02-02
pagecache: Replace reference to sled to that of pagecache

jeehoonkang:
Hi!  I just replaced a few mention of "sled" into that of "pagecache" in the pagecache crate.  Because maybe pagecache can be a standalone crate that can be used outside of the context of sled :)  Thanks!

### DIFF
--- a/crates/pagecache/src/config.rs
+++ b/crates/pagecache/src/config.rs
@@ -29,7 +29,7 @@ use io::LogReader;
 ///
 /// Read-only mode
 /// ```
-/// let _config = sled::ConfigBuilder::default()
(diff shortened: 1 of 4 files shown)


jeehoonkang:
Honestly I currently don't understand the error log, but it seems genuine: https://travis-ci.org/spacejam/sled/jobs/336377104

I don't think the filename changes in this PR are not likely the cause of the build failure, but I'm not 100% sure.

spacejam:
@jeehoonkang thanks for this!

Your change definitely didn't cause this bug! This appears to be a new crash that hasn't jumped out yet, exciting!

PR #218, opened 2018-02-02
pagecache: Refactor the periodic module

jeehoonkang:
I made a small PR that refactors the `periodic_flusher` module in `pagecache`.  `Log` periodically flushes its `Iobufs`.  I refactored this logic and made somewhat generic `Periodic` struct, which periodically calls the given callback until dropped.


crates/pagecache/src/io/periodic.rs
```diff
@@ -0,0 +1,51 @@
+use std::marker::PhantomData;
```
spacejam: Could you move this file to `crates/pagecache/src`? This looks great :)

jeehoonkang: I moved it :)

spacejam:
Awesome, thank you!

PR #221, opened 2018-02-04
pagecache: Add RAII-style and lambda-style measure

jeehoonkang:
The pagecache crate measure the time spent on running a piece of code for several times.  This PR introduces an RAII-style and lambda-style utility for measuring the time, which, IMHO, slightly improves the readability of the code.


spacejam:
Yeah, I like the style of what you wrote a lot more! thanks :)

PR #222, opened 2018-02-04
pagecache: Change Segment::lsn's type from Option<Lsn> to Lsn

jeehoonkang:
Currently `Segment::lsn`'s type of `Option<Lsn>`, but it seems it can never be `None`.  So I changed its type to `Lsn`.

I'm not sure it's the right thing to do: maybe `pagecache` has a plan to use `None` for `Segment::lsn` in a significant way.  Please let me know if it's the case!

### DIFF
--- a/crates/pagecache/src/io/segment.rs
+++ b/crates/pagecache/src/io/segment.rs
@@ -122,7 +122,7 @@ pub struct Segment {
     present: BTreeSet<PageID>,
     removed: HashSet<PageID>,
     deferred_remove: HashSet<PageID>,
-    lsn: Option<Lsn>,
+    lsn: Lsn,
     state: SegmentState,
 }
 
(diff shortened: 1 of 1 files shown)


jeehoonkang:
Hmm, definitely there's something I couldn't understand.  The CI build fails: https://travis-ci.org/spacejam/sled/jobs/337191874#L579   I don't have any clue of it..

jeehoonkang:
I'm sorry, `Segment::default()` initializes `lsn = None`, which means the segment is not used at all.  Closing it.

spacejam:
For context: the reason I went with an option here is because a segment can get freed or reallocated over time. I wanted to have a way of blowing up loudly when there are bugs in the lsn of the segment, because if this is wrong, it will cause tons of other problems elsewhere.

PR #225, opened 2018-02-06
pagecache: Inline droppers

jeehoonkang:
In this PR, `PidDropper` and `SegmentDropper` is inlined in the calls to `epoch::Guard::defer()`.  I think it's slightly more readable: you literally defers to call a function, rather than defers to drop an object.

### DIFF
--- a/crates/pagecache/src/io/segment.rs
+++ b/crates/pagecache/src/io/segment.rs
@@ -25,7 +25,7 @@
 //!    stable storage before we overwrite the
 //!    segment that may have contained the previous
 //!    latest stable copy of a page's state.
-//! 2. we use a `SegmentDropper` that guarantees
+//! 2. we use a `epoch::Guard::defer()` that guarantees
 //!    any segment that has been logically freed
 //!    or emptied by the `PageCache` will have its
 //!    addition to the free segment list be delayed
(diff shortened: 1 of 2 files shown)


crates/pagecache/src/io/segment.rs
```diff
@@ -564,8 +547,7 @@ impl SegmentAccountant {
             self.ordering.remove(&lsn);
         }
 
-        self.ordering = self.ordering
-            .clone()
+        self.ordering = mem::replace(&mut self.ordering, BTreeMap::new())
```
spacejam: I think if this was called frequently I would agree with this, but it's only called one time when the system is starting up, so I think `clone` is a little easier to follow. But let me measure what the cost might be for a pretty big database. If you have 4TB of storage, self.ordering has that / `io_buf_size` (8MB default), this will have 500,000 entries.

```rust
    #[bench]
    fn bench_add_two(bencher: &mut Bencher) {
        let mut btree = std::collections::BTreeMap::new();

        for i in 0..500000 {
            btree.insert(i, i);
        }

        bencher.iter(|| btree.clone());
    }
```
```
bench:  10,277,229 ns/iter (+/- 1,010,216)
```

So, 10ms for a one-time-cost on a pretty big database (on my laptop), but I think `clone` instead of `mem::replace` is more readable. I really want to prioritize readability right now for startup logic, which is already pretty confusing to follow. It's been one of the areas that has had the most bugs, which also makes me more hesitant to prioritize performance over readability here. To me, a tiny amount of readability is worth 10ms at startup, at least for now. Maybe we should revisit this in the future if we need to support use cases where databases are aggressively opened and closed, though.

spacejam:
I really like cleaning up the droppers structs! I agree that this is more readable. Thank you!

crates/pagecache/src/io/segment.rs
jeehoonkang: Oops!  I'm sorry I sneaked in some irrelevant change in this PR.  I'm reverting it.

Anyway, honestly I think `mem::replace()` is more readable, as it does what I can expect: take `self.ordering`, change it, and then reassign it to `self.ordering`.  On the other hand, `self.ordering.clone()` does what is not strictly necessary so I don't expect: cloning the ordering.

Issue #220, opened 2018-02-03
lurking corruption bug


spacejam:
the crash tests are rarely erroring on travis with reads hitting zeroed segments. a crash-related corruption bug lurks! kill it!!!!

spacejam:
ALICE may be helpufl for this, as it could arise from filesystem configuration differences between my laptop and the servers travis runs tests on.

spacejam:
had the first tree quickcheck result in a read on a zeroed out segment, unfortunately it is not deterministically reproducible
```
test quickcheck_tree_matches_btreemap ... FAILED
failures:
---- quickcheck_tree_matches_btreemap stdout ----
	thread 'quickcheck_tree_matches_btreemap' panicked at 'assertion failed: `(left == right)`
  left: `7295`,
 right: `0`: expected a segment header lsn that is divisible by the io_buf_size (10000) instead it was 4294967295', crates/pagecache/src/io/iterator.rs:117:8
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread 'quickcheck_tree_matches_btreemap' panicked at '[quickcheck] TEST FAILED (runtime error). Arguments: (OpVec { ops: [Del(21), Restart, Scan(251, 34), Del(171), Scan(246, 14), Get(148), Scan(250, 32), Set(134, 253), Cas(53, 246, 179), Del(160), Restart, Restart, Set(99, 65), Set(127, 198), Restart, Get(142), Get(221), Del(251), Get(37), Scan(60, 35), Scan(213, 9), Set(172, 13), Restart] }, 0, 0)
Error: "assertion failed: `(left == right)`\n  left: `7295`,\n right: `0`: expected a segment header lsn that is divisible by the io_buf_size (10000) instead it was 4294967295"', /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-0.6.1/src/tester.rs:179:27
```

spacejam:
```
thread 'test_crash_recovery_no_runtime_snapshot' panicked at 'assertion failed: `(left == right)`
  left: `24401216`,
 right: `24801216`', crates/pagecache/src/io/log.rs:160:17
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
[...]

PR #227, opened 2018-02-07
fix some test warnings

rleungx:
Fix some warnings when running `cargo test`.
Also there are a lot of operations like set, del called without check the result.


spacejam:
Thanks for fixing these warnings, @rleungx ! I believe the crash test failure is the same issue in https://github.com/spacejam/sled/issues/220 and I just need to sit down for a few hours while burning it in to get to the bottom of it with a good trace.

PR #229, opened 2018-02-07
Rip out cpuprofiler and env_logger

### LINKED ISSUE #224: rip out env_logger, cpuprofiler, global_init from crates/pagecache/src/lib.rs and callers
now that the pagecache is a little more split apart, we don't need pagecache to depend on these crates, or to call them. we should just remove the global_init function, and these two crates from any crates that pull them in (except stress2, which should keep env_logger)

jeehoonkang:
Closes #224.  The motivation behind this PR is explained in that issue :)


spacejam:
Thanks, @jeehoonkang :)

PR #231, opened 2018-02-08
implement Display for Error

### LINKED ISSUE #223: Display implementation for Error
It'd be nice to have a more friendly representation for printing errors.

What would you think of either using `failure`'s derive syntax to implement Display for Error, or implementing it manually? It could be implemented `where Actual: Debug`, and use the debug printing for `CasFailed`.

rleungx:
Closes #223.


spacejam:
Thanks, @rleungx !

PR #246, opened 2018-02-11
Remove prev from SegmentHeader

### LINKED ISSUE #230: pagecache: `SegmentHeader::prev` seems not used at all
`segment.rs` says:

```
//! 3. This case is solved again by having used
//!    <# io buffers> segments before reuse. We guarantee
//!    that the last <# io buffers> segments will be
//!    present, which means we can write a "previous
//!    log sequence number pointer" to the header of
//!    each segment. During recovery, if these previous
//!    segment Lsn pointers don't match up, we know we
//!    have encountered a lost segment, and we will not
//!    continue the recovery past the detected gap.
```

I assume that `SegmentHeader::prev: LogID` is the "previous log sequence number pointer".  However, it seems it is not used at all, except for de/serialization.  I'd like to ask I understood it correctly.  Thanks!

jeehoonkang:
This change was discussed in #230.  Closes #230.

@spacejam Thanks for guiding me on this issue!

### DIFF
--- a/crates/pagecache/src/io/segment.rs
+++ b/crates/pagecache/src/io/segment.rs
@@ -56,11 +56,10 @@
 //! 3. This case is solved again by having used
 //!    <# io buffers> segments before reuse. We guarantee
 //!    that the last <# io buffers> segments will be
-//!    present, which means we can write a "previous
-//!    log sequence number pointer" to the header of
-//!    each segment. During recovery, if these previous
-//!    segment Lsn pointers don't match up, we know we
-//!    have encountered a lost segment, and we will not
+//!    present, from which can deduce the "previous log
+//!    sequence number pointer". During recovery, if these
+//!    previous segment Lsn pointers don't match up, we know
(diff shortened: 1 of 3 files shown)


crates/pagecache/src/io/segment.rs
```diff
@@ -56,11 +56,10 @@
...
 //!    that the last <# io buffers> segments will be
-//!    present, which means we can write a "previous
-//!    log sequence number pointer" to the header of
-//!    each segment. During recovery, if these previous
-//!    segment Lsn pointers don't match up, we know we
-//!    have encountered a lost segment, and we will not
+//!    present, from which can deduce the "previous log
+//!    sequence number pointer". During recovery, if these
+//!    previous segment Lsn pointers don't match up, we know
+//!    we have encountered a lost segment, and we will not
```
spacejam: 👍

spacejam:
Hell yeah! Nice work :) I'm really happy to have some company in these correctness-critical and tricky to reason about parts of the system!

PR #247, opened 2018-02-11
Cleanup minor things

jeehoonkang:
- I removed unnecessary `preodic_flusher.rs`, which I forgot to remove in #218.  Sorry!
- I removed an unnecessary duplicated ifs in `reader.rs`.
- I removed unnecessary cleanup code for `SegmentAccountant::ordering`.  It is already cleaned up just before.


spacejam:
Thanks!

PR #265, opened 2018-02-20
prefix encoding. closes #264

### LINKED ISSUE #264: store keys as [low key shared prefix bytes, remainder]
Every `Tree` page stores a low key. We pay a high cost for key storage. We can instead treat the first byte as a count from 0-255 of the bytes from the low key that each key shares a common prefix with.

Low key: cat
cat -> [\x03]
ca -> [\x02]
cab ->[\x02b]
cvar -> [\x01var]
zigbee -> [\x00zigbee]

spacejam:


nrc:
Question about `prefix_cmp`: iiuc, a prefix-encoded key has as it's first element the length of the prefix which is elided, and then the rest of the key is the literal suffix. I assume that an invariant of `prefix_cmp` is that the two keys are prefix-encoded against the same prefix. However, even so it seems to me that comparing the first elements is incorrect because if the two keys don't take the same size prefix, then we can't compare them without the prefix - each element in the prefix might be greater or less than an element in the other key.

Furthermore, it seems that like `prefix_decode`, `prefix_cmp` should not operate on empty keys.

So, am I misunderstanding prefix encoding or prefix cmp? Thanks!

spacejam:
@nrc I just spent some time beefing up some tests that "should" have caught this obvious bug before remembering the invariant that holds here which makes it all OK: we always use a prefix that is a lexicographic lower bound for all keys in a bw tree node (specifically, we just use the node low bound as the prefix).

In any case, thanks for the encouragement to beef up some tests and improve documentation :]

nrc:
Ah, I see, thanks for the explanation!

PR #281, opened 2018-03-01
rearrange subcrates a bit

spacejam:


README.md
```diff
@@ -59,9 +59,7 @@ tree.del(&k);
 * MVCC, transactions, merge operators and snapshots provided via a higher-level `Db` versioned-key interface
 * custom merge operators a la RocksDB
 * form the iron core of a [linearizable store](https://github.com/spacejam/rasputin) and a [flexible location-agnostic store](https://github.com/spacejam/icefall)
-* SQLite, MySQL, Postgres back-end plugin support
```
jeehoonkang: I'm just wondering, is this postponed, or cancelled?

spacejam: definitely not cancelled, but I'm trying to focus my energy on fewer things right now. I'm totally up for working with other people to help get that done in the short / medium term, but for now I want to focus on getting the core of this system into a stronger place, while just keeping an eye on SQLite, MySQL, and Postgres requirements for storage engines so we don't make things harder than necessary to eventually support them.

PR #297, opened 2018-03-08
put created files into a subdirectory

### LINKED ISSUE #240: create a subdirectory based on path, and use standard storage file basenames
now that we are creating several files (`path.conf`, `path.db`, `.shapshot.00000XXXX...` it is cleaner to just put all of the state files into a subdirectory. Then we don't need to deal with `OsStr` all over the place in ugly ways internally.

rleungx:
Resolves #240.

### DIFF
--- a/crates/pagecache/src/config.rs
+++ b/crates/pagecache/src/config.rs
@@ -1,7 +1,7 @@
 use std::fmt::Debug;
 use std::fs;
 use std::ops::Deref;
-use std::ffi::{OsStr, OsString};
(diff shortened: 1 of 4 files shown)


rleungx:
I made some changes. 
PTAL @spacejam

spacejam:
@rleungx nice! thank you very much for helping with this! I'll cut a new version soon that contains your work :)

spacejam:
just cut pagecache `0.4.2` and sled `0.15.13` with your changes

Issue #145, opened 2017-09-21
32-bit ARM compatibility


spacejam:
- [x] add test to make sure that the system never wraps by writing to an offset > 2^32
- [x] add additional concurrency verification tooling, as ARM has a weaker memory model than x86_64 (which makes it a great test platform in general)

rubdos:
Maybe add ARM target to CI? I'm stumbling on a few bugs, so expect PR's.

spacejam:
@rubdos hey! sorry for the slow response, just started a new job and have been getting used to my new schedule. yeah, this is something I've been thinking a lot about. I think the correct solution here is to use double-wide LL/SC operations on 32-bit ARM architectures that support it, but this will require writing a little bit of C I believe for that target.

rubdos:
There's no rush for me currently, but for me it's currently the blocker to port my application to Android! Not that I'm planning that anytime soon though. So take your time to start your new job; I wish you the best of luck!

--

I wonder though, aren't those ll/sc operations what AtomicI64 would use?

Is it, for you, an idea to have a `nightly` feature in sled, which uses AtomicI64 for the LSN? Then it's easily switched when AtomicI64 gets stabilized.

spacejam:
@rubdos I would be OK with conditional compilation of a nightly-only feature for 32-bit targets, so long as 64-bit users can run it on stable.

rubdos:
Roger that, I'll implement that the next few hours :-)

spacejam:
@rubdos that's super exciting, thanks for letting me know! At work we actually just got a batch of raspberry pi's, and I was thinking about running one in more or less a constant loop with the `lock_free_delays` feature set to get concurrency bugs to pop out. ARM is nice for this because of the weaker memory model, which should have a higher chance of causing bugs to pop out.

rubdos:
That's cool too! If you want a bit faster ARM machine, you should consider the (aarch64) Odroid XU4. That thing is an octa core 64 bit ARM SBC! Might be useful too. I don't have one currently, but I'm planning to buy one for the same purpose: debugging concurrency bugs :-)

spacejam:
@rubdos I just ordered one :D looking forward to trying to get this thing melt hahaha

rubdos:
Nice, good luck melting it. I am *quite* confident that you won't succeed ;-)

rubdos:
FYI: I had a few crashes on an x86 Android emulator (not even ARM), so seemingly some troubles with 32 bit devices.

[...]

PR #300, opened 2018-03-19
Compilation for 32-bit architectures on nightly

rubdos:
Discussion points:

- [x] Is a global `nightly` feature good in this case? Should I give it a better name?
- [x] I didn't run the tests on an ARM 32-bit; would you want me to test that?
- [x] `pagecache::Lsn` is now of type `i64` when the `nightly` feature is enabled.

Let me know if you want me to change stuff! :-)

Issue #145 for reference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,18 @@ matrix:
       env:
         - TEST=crash
         - RUST_BACKTRACE=1
+    - rust: nightly-2018-01-02
+      env:
+        - TEST=cross-build
+        - TARGET=i686-unknown-linux-gnu
+
+install:
+  - curl https://sh.rustup.rs -sSf |
+    sh -s -- -y --default-toolchain $TRAVIS_RUST_VERSION
+  - if [ -n "$TARGET" ]; then
+        rustup target add $TARGET;
+    fi
+  - source ~/.cargo/env
 
 script:
   - bash -c 'case "$TEST" in
@@ -37,6 +49,11 @@ script:
                  ;;
                crash)
                  cargo test test_crash_recovery --release --features="check_snapshot_integrity" -- --nocapture
+                 ;;
+               cross-build)
+                 echo "https://github.com/rust-lang/cargo/issues/4753";
+                 pushd crates/sled; cargo build --target $TARGET --features=nightly; popd;
+                 pushd crates/pagecache; cargo build --target $TARGET --features=nightly; popd
                  ;;
                *)
                  cargo check;
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ assert_eq!(tree.get(&k), Ok(Some(vec![4])));
   operations per second with mixed workloads, and 7 million/s
   for read-only workloads on tiny keys. this will be improving 
   dramatically soon!
+* 32 bit architectures [require Rust nightly with the `nightly` feature enabled](https://github.com/spacejam/sled/issues/145).
 
 # contribution welcome!
 
--- a/crates/pagecache/Cargo.toml
+++ b/crates/pagecache/Cargo.toml
@@ -16,6 +16,7 @@ lock_free_delays = ["rand"]
 failpoints = ["fail"]
 no_metrics = ["historian/bypass"]
 no_logs = ["log/max_level_off"]
+nightly = []
 
 [dependencies.historian]
 version = "3.0"
--- a/crates/pagecache/src/io/iobuf.rs
+++ b/crates/pagecache/src/io/iobuf.rs
@@ -1,5 +1,9 @@
 use std::sync::{Condvar, Mutex};
-use std::sync::atomic::{AtomicBool, AtomicIsize, AtomicUsize};
+use std::sync::atomic::{AtomicBool, AtomicUsize};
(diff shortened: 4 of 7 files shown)


rubdos:
FYI; I ran a 32 bit ARM executable on my Moto Z Play (which admittedly supports aarch64):

```
running 3 tests
test tree::bound::test_bounds ... ok
test tree::prefix::test_prefix ... ok
test tree::prefix::test_prefix_cmp ... ok
```

spacejam:
Sorry for being so slow with this! We can actually just use `#[cfg(nightly)]` and `#[cfg(not(nightly))]` instead of defining a new feature.

spacejam:
The one other thing, is that maybe we want to generally use AtomicI64 when `#[cfg(target_pointer_width = "32")]`, so we never even mention nightly or not, just try to pull in integer atomics and use them when target_pointer_width is 32. Then only define AtomicLsn as AtomicIsize when target_pointer_width = "64". What do you think about that approach?

rubdos:
> Sorry for being so slow with this! We can actually just use #[cfg(nightly)] and #[cfg(not(nightly))] instead of defining a new feature.

Didn't know that was a thing, will change.



> The one other thing, is that maybe we want to generally use AtomicI64 when #[cfg(target_pointer_width = "32")], so we never even mention nightly or not, just try to pull in integer atomics and use them when target_pointer_width is 32. Then only define AtomicLsn as AtomicIsize when target_pointer_width = "64". What do you think about that approach?

That means that it will still give a strange error on 32 bit non-nightly. I'll have it throw some more meaningful message for that.

This also means that at a certain point in time, this construct will "just work" on stable, which means there'll be less incentive to change everything to `i64`, won't it? But I think I agree that this is a cleaner solution.

rubdos:
> This also means that at a certain point in time, this construct will "just work" on stable, which means there'll be less incentive to change everything to i64, won't it? But I think I agree that this is a cleaner solution.

It won't, I added this compile error:

```
error: 32 bit architectures require a nightly compiler for now.
               See https://github.com/spacejam/sled/issues/145
  --> crates/pagecache/src/lib.rs:9:1
   |
9  | / compile_error!("32 bit architectures require a nightly compiler for now.
10 | |                See https://github.com/spacejam/sled/issues/145");
   | |_________________________________________________________________^

error: aborting due to previous error

error: Could not compile `pagecache`.

To learn more, run the command again with --verbose.
```

rubdos:
If you ack this one, let me squash it before you merge :-)

rubdos:
> Sorry for being so slow with this! We can actually just use #[cfg(nightly)] and #[cfg(not(nightly))] instead of defining a new feature.

Actually, I cannot seem to find any documentation beyond [this thread](https://internals.rust-lang.org/t/setting-cfg-nightly-on-nightly-by-default/1893) on this, are you sure this exists? Doesn't seem to work here.

rubdos:
FYI: I just found out that `std::sync::atomic::hint_core_should_pause()` doesn't exist in nightly anymore, in [`iobuf.rs`, line 1195](https://github.com/spacejam/sled/blob/71a7fe528a2c896686343a8577a0d9cccc8d3660/crates/pagecache/src/io/iobuf.rs#L1183), and the `cfg(nightly)` you have never hits the compiler. You can try it by having `compile_error!("This gets compiled")` in the second branch, with `not(nightly)`, and then compile with a nightly compiler.

spacejam:
> cfg(nightly) you have never hits the compiler.

my bad! that codepath is only triggered during tests, which I assumed were running some of in nightly, but that was a bad assumption. I think I got the idea from that thread you linked, and just tried it and it didn't blow up until you found it just now because I wasn't properly exercising that codepath.

spacejam:
perhaps something from this travis file can be used to get i686 building on travis? https://github.com/japaric/cross/blob/master/.travis.yml

rubdos:
> perhaps something from this travis file can be used to get i686 building on travis? https://github.com/japaric/cross/blob/master/.travis.yml

Right, I didn't check whether the thing I pushed actually worked on travis >.<

Let me have a check :-)

rubdos:
Hehe, that works. Would you like some more (32 bit) archs to be added?

spacejam:
> Should I add a note to the README.md saying that 32 bit requires --features=nightly and a nightly compiler? :-)

That would be wonderful! I'm really excited to have support for these architectures now :)

rubdos:
Heh, you got yourself a race condition in the Travis build of the README change. Lol.

spacejam:
@rubdos a gift from the test gods!

These failpoint property tests reveal a steady stream of error handling inconsistencies, it's great, but also feels like an endless stream of bugs sometimes, and can be a bit time consuming to fix. I believe they are relatively low priority compared to a couple other issues right now, so it might be a couple weeks before this is green for very long test runs, but most of the time it passes.

I'll rerun the test and merge your changes, and cut a new release with them :)

spacejam:
@rubdos the latest versions of pagecache and sled have these changes in! thanks again for your good work on this  :)

rubdos:
Hooray! Thanks for the feedback! I am looking forward to the next patch! :D

PR #305, opened 2018-04-05
 64-bit target with `nightly` feature doesn't compile anymore.

### LINKED ISSUE #304: 64-bit target with `nightly` feature doesn't compile anymore.
Since #300, using `nightly` feature with a 64-bit target doesn't compile anymore...

```
   Compiling pagecache v0.4.3 (file:///home/rsmet/src/sled/crates/pagecache)
error[E0308]: mismatched types
    --> crates/pagecache/src/io/segment.rs:1212:18
     |
1212 |         max_lsn: std::i64::MAX,
     |                  ^^^^^^^^^^^^^ expected isize, found i64

error: aborting due to previous error
```

rubdos:
Apparently I missed a case, where `--features=nightly` and target is a 64 bit arch...

I'm wondering whether it would be better to use `#[cfg(nightly)]` everywhere, and only use `#[cfg(target_pointer_width="32")]` for the error message. That way, the i64 type gets thoroughly tested with the nightly flag, independently from whether we're targetting 32 bit...

(fixes #304)

### DIFF
--- a/crates/pagecache/src/io/segment.rs
+++ b/crates/pagecache/src/io/segment.rs
@@ -1208,9 +1208,9 @@ pub fn raw_segment_iter_from(
 
     Ok(LogIter {
         config: config.clone(),
-        #[cfg(feature = "nightly")]
+        #[cfg(target_pointer_width = "32")]
         max_lsn: std::i64::MAX,
-        #[cfg(not(feature = "nightly"))]
+        #[cfg(target_pointer_width = "64")]
         max_lsn: std::isize::MAX,
         cur_lsn: SEG_HEADER_LEN as Lsn,
         segment_base: None,


daboross:
Could we maybe do `and(target_pointer_width, nightly)` and `not(and(target_pointer_width, nightly))`?

rubdos:
Sorry @daboross, but I think what you wrote *might* make some sense for discussion, so I hope you don't mind me digging up your comment:

> Could we maybe do `and(target_pointer_width, nightly)` and `not(and(target_pointer_width, nightly))`?

While `and` (resp. `or`) doesn't exist, there exists `all` (resp. `any`). I think it would make sense to use `any` in this case, although generating a convenience flag somehow would be nice too (such that we can do `cfg(use_i64)`). Not sure whether that last part is possible in Rust.

Using `any(feature = "nightly", target_pointer_width = "32")`, we can guarantee that `i64` is used in both the `nightly` and the `target_pointer_width = "32"` case.

daboross:
@rubdos no problem!

I deleted it since I realized my thought was incomplete, and didn't yet have time to research the whole idea.

I was thinking if we just do `#[cfg(target_pointer_width = "32")]`, then the user might get confusing error messages if they're on 32bit and don't have `nightly` selected since it would depend on the nightly feature. But, we have the nice compile error statement in `pagecache` which should deal with that!

`#[cfg(any(target_pointer_width = "32"), feature = "nightly")]` is different but sounds like a good idea! It would mean depending projects get errors if they're compiling for 64bit, but with nightly feature, and something about our use of nightly for this breaks. I think that's a good thing?

rubdos:
> #[cfg(any(target_pointer_width = "32"), feature = "nightly")] is different but sounds like a good idea! It would mean depending projects get errors if they're compiling for 64bit, but with nightly feature, and something about our use of nightly for this breaks. I think that's a good thing?

I'm not sure. Since we have that nice compile error, we can really do anything at this point; it doesn't really matter.

I would though aim to use `i64` in any nightly case, not only when we're on 32 bit... But I guess we need @spacejam to comment on that! :-)

spacejam:
I think we're just going to need to have clear documentation for 32-bit users to know they need to use

PR #307, opened 2018-04-27
pagecache: get rid of several unwraps

lucab:
This removes several unwrapping points, handling options via match
statements instead. It also bubbles up a few additional error cases.

### DIFF
--- a/crates/pagecache/src/io/page_cache.rs
+++ b/crates/pagecache/src/io/page_cache.rs
@@ -287,9 +287,12 @@ impl<PM, P, R> PageCache<PM, P, R>
 
     /// Return the recovered state from the snapshot
     pub fn recovered_state(&self) -> Option<R> {
-        let mu = &self.last_snapshot.lock().unwrap();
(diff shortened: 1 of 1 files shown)


spacejam:
Thanks @lucab :D

lucab:
:+1: 
There are still quite a few panics and unwraps, it would be nice if you could mark the ones you want to retain long-term (if any), and add TODOs for the spurious ones.
I don't have any specific usage for sled/pagecache at the moment, but this kind of cleanups are perfect for killing [idle times](https://xkcd.com/303/) in a useful way.

spacejam:
Maybe at some point today I'll sweep through and change all fail-stop
worthy ones from unwrap to expect. Earlier in the project everything was
fail stop, but especially after the failpoint tests were added the goal is
to encapsulate all failures that can result from io issues.

spacejam:
been going through a bunch of these, about 100 left still lol... but going through and being more thorough has already caught a few interesting bugs

PR #316, opened 2018-05-27
Fix or ignore failing tests

rtsuk:
I'm not sure you'll want to commit these changes, but this was what I had to do to get a test run with no failures.

A number of them are because there is no concept of working directory in Fuchsia yet. The others I had not idea why they were failing.

### DIFF
--- a/crates/pagecache/src/config.rs
+++ b/crates/pagecache/src/config.rs
@@ -114,7 +114,7 @@ impl Default for ConfigBuilder {
         let tmp_path = format!("/dev/shm/pagecache.tmp.{}", salt);
 
         #[cfg(not(target_os = "linux"))]
-        let tmp_path = format!("pagecache.tmp.{}", salt);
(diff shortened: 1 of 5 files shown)


spacejam:
I think the best option is to change the `#[ignore]` attributes to `#[cfg(not(target_os = "fuchsia"))]` for now. I'll add this now and merge!

spacejam:
@rtsuk what's the best way to get fuchsia up and running so I can debug the failures?

rtsuk:
The [getting started instructions](https://fuchsia.googlesource.com/docs/+/master/getting_started.md) will walk you through it. My only recommendation is to use `fx set-layer garnet` rather than `topaz`, as it is a lot less source to download and compile. At the end you'll have a system image and compatible build of Qemu that can use with [fargo](https://fuchsia.googlesource.com/fargo/) to build and test sled.

If you try it and have any difficulties, let me know.

spacejam:
@rtsuk awesome, thanks!

PR #323, opened 2018-07-04
Fix model doctests compilation

mbr:
A few small fixes that make the doctests compile (tested using Rust 1.27.0).


spacejam:
Thanks!

PR #333, opened 2018-08-09
support for off-log external blobs. closes #198.

### LINKED ISSUE #198: punt huge values to external files
a current limitation of the lock-free log is getting values larger than an io buffer segment written to disk

spacejam:
this addresses one of the larger pain points that limited use cases of sled. now you can store large values, which are placed in the `blobs` subdirectory, indexed by original lsn. at attempt is made to keep them linearized with the log, by reading all post-checkpoint blobs and verifying their crc before advancing the associated log index during recovery. any blobs that have a higher lsn than the recovered log are deleted, as they are from an unrecoverable future. if there is interest maybe they can be moved to a "lost and found" directory instead, though.

these changes will be available in sled `0.15.18` and pagecache `0.4.6`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ rust:
 
 matrix:
   include:
-   - rust: nightly-2018-09-13
+    - rust: nightly-2018-09-13
       env:
         - TEST=tsan
         - RUSTFLAGS="-Z sanitizer=thread"
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,20 +9,20 @@ rust:
 
 matrix:
   include:
-    - rust: nightly-2018-05-03
+    - rust: nightly-2018-08-04
       env:
         - TEST=tsan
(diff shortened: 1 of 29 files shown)


spacejam:
remaining items before this is mergeable:

- [ ] pagecache aware of whether a page fragment is resident in the log or blob area, so that it can optimize rewriting pointers instead of creating a new massive blob file
- [ ] pagecache uses epoch dropper for replaced and deleted external pages that guarantees removals of actual blob files occur after the log is guaranteed to be stable and reachable during recovery (barring corruption)
- [ ] during recovery, replay all removals of blobs to remove any files that should have been deleted but the process crashed before it happened

bouncing between this and making notes about implementing anti-caching on a plane now

spacejam:
- [ ] when replacing a page with more than one pointer, and some of the pointers are External, use a segment-specific dropper that defers all rm's until that segment transitions to Inactive

PR #340, opened 2018-09-23
More variety to the keys used in Tree-BTreeMap equivalence test

spacejam:


crates/sled/src/tree/prefix.rs
```diff
@@ -33,7 +33,11 @@ pub fn prefix_decode(prefix: &[u8], buf: &[u8]) -> Vec<u8> {
     ret
 }
 
-pub fn prefix_cmp(a: &[u8], b: &[u8]) -> Ordering {
+// NB: the correctness of this function depends on
+// the invariant that the prefix is ALWAYS lexicographically
+// Less than or Equal to the keys that have been encoded
+// using it. Otherwise this comparison would make no sense.
+pub(crate) fn prefix_cmp(a: &[u8], b: &[u8]) -> Ordering {
```
spacejam: @nrc would this have prevented the confusion, or do you think I should make it even more clear?

PR #352, opened 2018-10-16
zero copy readpath pagecache. closes #349

### LINKED ISSUE #349: Always fix-up pagecache stacks on read, and return pointer for zero-copy reads
Currently a huge amount of time is spent performing a copy of the PageFrag returned by `PageCache::page_in` when a `Materialized` page is returned. We don't need to make a copy of this if we always fix-up the stack on reads, and return a pointer to the new fixed-up head. The returned PageFrag is bound to the epoch guard's lifetime anyway, so we get nothing by making a whole copy of it. 

Right now if the fix-up fails, we don't retry, but still return the stale read to the caller. In most cases this is fine, but if the caller is interested in later performing a link or replace based on the results of a get, then they are going to be wasting all of the work that they do from this point on.

Getting rid of the copy will let us offer an end-to-end zero-copy read for sled for users who wish to use a slightly more complex API involving grabbing an EBR guard before access, which is pretty cool.

Steps:
* `PageCache::page_in`
  * if the page isn't Free, Allocated, or Unallocated, we should spin until a MergedResident is at the head of the page stack, through our efforts or those of other threads
* `PageGet::Materialized` can drop the first item of `PageFrag`
* `PagePtr` can implement `Deref` for its inner item.
  * This is guaranteed to be a fully-merged node, because we fixed it up in `page_in`

Anything that relied on results from `PageCache::get` now needs to do their own clones if it's necessary.

spacejam:


crates/pagecache/src/pagecache.rs
```diff
@@ -99,7 +112,7 @@ where
     /// This page contains data and has been prepared
     /// for presentation to the caller by the `PageCache`'s
     /// `Materializer`.
-    Materialized(PageFrag, PagePtr<'a, PageFrag>),
+    Materialized(&'a PageFrag, PagePtr<'a, PageFrag>),
```
spacejam: this is the key performance result of this PR

PR #354, opened 2018-10-16
Fix `USAGE` string in stress2

nrc:
kv-len was present in the long help message but not the short one, so was ignored by DocOpt


spacejam:
Thanks!

PR #358, opened 2018-10-18
Optimise prefix comparison

### LINKED ISSUE #285: optimize prefix comparison
we're allocating like crazy while traversing the tree and comparing keys. we should never allocate during the traversal of a path that is already in memory and doesn't require merging.

nrc:
Closes #285

Output from `stress2`:

```
master
traverse |        2.5 |       73.9 |      129.3 |      224.1 |     8537.7 |     699934 |     27.632

prefix-opt
traverse |        1.7 |       80.8 |      135.9 |      206.9 |     1045.5 |     649053 |     27.825
```

There's not much change in the smaller columns, but since that's only a handful of us, I assume it's not significant. The max is improved by more than 8x. I don't know how to interpret the 'sum' column.

Flamegraphs:

master:
master

prefix-opt:
prefix-opt

These flame graphs show the same view of stress2. On master, there is significant time spent in `prefix_decode` and jemalloc. On the prefix-opt branch, both have disappeared (they are not even present on hovering the small items on the far right, and nor is any other tree comparison function.

r? @spacejam

### DIFF
--- a/crates/sled/src/lib.rs
+++ b/crates/sled/src/lib.rs
@@ -64,7 +64,7 @@ use self::bound::Bound;
 use self::data::Data;
 use self::frag::{ChildSplit, ParentSplit};
 use self::node::Node;
-use self::prefix::{prefix_cmp, prefix_decode, prefix_encode};
+use self::prefix::{prefix_cmp, prefix_decode, prefix_encode, prefix_cmp_encoded};
 
 pub(crate) use self::frag::Frag;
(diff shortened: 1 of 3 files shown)


spacejam:
oooo this is wonderful :D great job!!! so excited to run some benchmarks on this in combination with the recent zero-copy readpath stuff!!!

pushed a tiny change to make the new `prefix_cmp_encoded` `pub(crate)` to make the `rust_2018_idioms` lint happy.

PR #361, opened 2018-10-21
fix compilation error on windows

ozgur11235813:
I couldn't compile on windows due to an error that said: "no uptime() defined". This fixes compilation but I couldn't run the tests on windows because of similar problems.

### DIFF
--- a/crates/pagecache/src/config.rs
+++ b/crates/pagecache/src/config.rs
@@ -4,7 +4,7 @@ use std::io::{Read, Seek, Write};
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{
-    AtomicPtr, AtomicUsize, Ordering, ATOMIC_USIZE_INIT,
+    AtomicPtr, AtomicUsize, Ordering,
 };
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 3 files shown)


spacejam:
`uptime` is a function defined internally to the `pagecache` crate in this repo. Could you paste your compiler output?

In general I'm opposed to adding new dependencies, especially for something as simple as this.

spacejam:
Ahh I see what the issue is, uptime is private and just needs to be pulled into the pagecache `lib.rs`.

crates/pagecache/src/config.rs
```diff
@@ -95,20 +95,10 @@ unsafe impl Send for ConfigBuilder {}
...
-        static SALT_COUNTER: AtomicUsize = ATOMIC_USIZE_INIT;
-
-        #[cfg(unix)]
-        let salt = {
-            let pid = unsafe { libc::getpid() };
-            ((pid as u64) << 32)
-                + SALT_COUNTER.fetch_add(1, Ordering::SeqCst) as u64
-        };
-
-        #[cfg(not(unix))]
```
spacejam: probably better to leave the unix counter as it is. We spin up large numbers of instances concurrently on some testing systems and uptime is not monotonic across processes. Even though the chance of a collision is pretty small, it's better to be correct to totally rule out false negative test results caused by multiple processes operating on the same data concurrently.

spacejam:
Thanks for changing that! I think once the unix salt is set back to the way it was, this is good to go!

ozgur11235813:
no. thank you :D

PR #362, opened 2018-10-21
model: Add ability to pass config to linearizable!

vorner:
* Fix some doctests on the way.
* Let the default (config-less) mode follow the PROPTEST_CASES
  environment variable if specified.


spacejam:
Awesome, thanks! I'm going to be sweeping through and releasing new versions of all of the crates in here shortly, so this will be on crates.io soon!

spacejam:
Awesome, thanks!

PR #366, opened 2018-10-23
Binary search least-upper-bound while traversing indexes 

spacejam:
This speeds up tree traversal at fanouts > a dozen or so


spacejam:
Pushed changes that include a fix to determine the actual size of an item sitting in cache.

spacejam:
probably just unearthed a cache-eviction-related race condition by fixing the way eviction works

spacejam:
This is a durability bug that has to do with node splits, rather than caching. Before, we were never actually causing node splits in the crash test because the splits were based solely on node fanout, and we set the same number of keys. Creating a separate issue to handle this.

PR #370, opened 2018-10-24
add no_inline feature to pagecache

### LINKED ISSUE #367: add no_inline feature to pagecache and sled that disables all inlining hints for profiling purposes
There are a few parts of the codebase where we use `#[inline]` and `#[inline(always)]`.

Occasionally, I'm curious how much time actually gets spent in these sections, but they are hidden from flamegraphs/cachegrind/callgrind due to being inlined into their calling functions.

All calls to `#[inline(always)]` should be changed to `#[cfg_attr(not(feature = "no_inline"), inline)]` so we can get better insight into their true costs on flamegraphs.

ozgur11235813:
closes #367


spacejam:
Looks good! Nice job :D

PR #377, opened 2018-10-28
simplify config. Closes #326.

### LINKED ISSUE #326: Moving the database on disk means we cannot open the database anymore.
We currently check the generated config against a serialized copy on disk here:
https://github.com/spacejam/sled/blob/master/crates/pagecache/src/config.rs#L445

Since the config also stores the path in `Config.inner.path`, this means that if we try to move the database we will fail the verify check.

Do you have any thoughts here on how to loosen up this constraint a bit?

*for context, I'm trying to replicate a sled database across a few devices, the database will most likely not be in the same place on each device since we have no guarantee that even the same operating system will be used*

**sled version:** 0.15.17
**rustc version:** 1.28.0-nightly (01cc982e9 2018-06-24)
**operating system:** macOS 10.13.6
**logs:**
```
SledGeneric(Unsupported("changing the configuration between usages is currently unsupported"))
```

spacejam:
adds a default tree open method
allows databases to be moved around now
fewer configuration options
add sloped cleanup thresholds that should drive 10% more cleanup efforts towards lower segments to keep db's leaner

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,9 @@ An alpha modern embedded database.
 ```rust
 extern crate sled;
 
-use sled::{ConfigBuilder, Tree};
+use sled::Tree;
 
-let config = ConfigBuilder::new()
-  .path(path)
-  .build();
-
-let tree = Tree::start(config).unwrap();
+let tree = Tree::start_default(path)?;
 
 // set and get
 tree.set(k, v1);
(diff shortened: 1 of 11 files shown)


spacejam:
* fixed bug with segment threshold skew calculation
* added another condition var which pretty much eliminates all needless spinning in reservation logic

PR #389, opened 2018-11-01
Add keyspace parameter to stress2

spacejam:


spacejam:
This lets us go against [leveldb](https://github.com/google/leveldb#performance) and easily destroy it for read performance with a laptop. For writes, this highlights how our copies on the writepath are really hurting us.

We can probably already beat [rocksdb](https://github.com/facebook/rocksdb/wiki/RocksDB-In-Memory-Workload-Performance-Benchmarks) for reads while sustaining writes as well, but we need to run a matching workload on a 32 thread machine.

PR #396, opened 2018-11-02
examples: make playground compatibility with v0.16

sunisdown:


spacejam:
Thanks for cleaning this up :D

Issue #393, opened 2018-11-01
Cleanup clippy suggestions


spacejam:
It has been way too long since we've swept through clippy suggestions and applied them. There is currently a lot of low-hanging fruit in here for helping to clean up the codebase. Any help would be quite appreciated :]

spacejam:
Also, even a single clippy warning cleanup would be a valuable contribution. This is definitely not an all-or-nothing item :)

PR #412, opened 2018-11-05
fix some clippy warnings

ozgur11235813:
related to #393


ozgur11235813:
@spacejam what do you think about "this argument is passed by value, but not consumed in the function body" warnings, can they be fixed. There are a lot of them.

spacejam:
@ozgur11235813 thanks for cleaning this up! 

re: consumed warnings, please don't change the ones in the public functions in `tree.rs` as I'm going to optimize the cloning away at some point, and we will consume them then.

ozgur11235813:
you can merge this I will continue another time

spacejam:
:+1:

Issue #411, opened 2018-11-04
multithread write seems to lost data


WenjunHuang:
Bug reports must include:
1. sled version
0.16.2

Hi,I wrote a simple multi-threaded write / read test for sled in [https://github.com/WenjunHuang/SledMultiThreadTest](https://github.com/WenjunHuang/SledMultiThreadTest)

I found that when i first used multi-threaded to write some data to sled and then __restart__ the program to start as multi-thread read, some data will lost.
But if i first write then read(without restart the program),everything was ok.
And If i used single thread write and read, everything was ok whether i restart the program after write or not.

spacejam:
The main issue should be fixed, so you can remove your `Cargo.lock` or just bump up pagecache to `0.7.3`!

This was a pretty lame bug, but the issues in #414 are going to significantly improve long-term reliability.

PR #414, opened 2018-11-05
[WIP] debug concurrent data loss

spacejam:
Live-streamed efforts to address #411


spacejam:
Bug reproduced in failing test.

Observations:
* happens even when no segments are reused, as the failing test does not reuse segments
* happens without a restart when a low cache size is specified on a mixed workload with stress2, although this does reuse segment
* when configured to use 1 io buffer, the system locks up
* when configured to use 3 io buffers, sometimes the same buffer writes twice in a row
* 2 threads doesn't easily reproduce
* 3 threads results in 1/6th-0 data loss at 500 byte splits 
* 10 threads results in 20-80% loss at 500 byte splits
* 5 threads: slightly less than 50% data loss when configured to split nodes to be 2 wide
* 5 threads: ~20-75% data loss with 500 byte splits
* 5 threads: 50-70% loss with 250 byte splits

there may be multiple issues. here are some suspects to rule out and add tests that try to catch, should they happen in the future:

1. some segments may not be recovered during restart (wouldn't explain stress2 runtime crash)
2. some writes may not ever hit the log
3. some writes may be overwritten improperly
4. during recursive node splits, some keys may be dropped
5. node compactions may race with caps / replaces
6. race during root hoist

spacejam:
Added a pretty cool test that uses linux realtime priorities + quickchecked operations + tree restarts to get something kind of like what the erlang paid quickcheck world is able to do with PULSE. It's failing when run in combination with the concurrent event log, which validates a few desired properties.

Taking a long time to shrink failures with the naive strategies, but this is a really exciting test strategy :]

spacejam:
Main bug isolated and addressed: in the IO buffer, when we seal a segment but there is enough space left over to fit at least another message header, we need to fill the remainder with a special message that signifies the buffer has been correctly filled. However, there was a race where another thread could write the buffer before the sealing thread was able to fill the remainder. This is fixable by either requiring the writer thread to perform the buffer padding.

In the process of scrutinizing different areas of the codebase, I've found a few other race conditions, and I'll incrementally fix issues in separate PR's.

spacejam:
putting fixes so far into pagecache 0.7.3, via #417

spacejam:
last known bug, which is quite minor, is fixed in #424

spacejam:
All tests, refactors and fixes are now merged in their own prs. Closing.

PR #422, opened 2018-11-12
Update methods to accept an AsRef<[u8]> type instead of a Vec<u8> 

Kerollmops:
This pull request change the `Tree` methods to accept a `&[u8]` instead of a `Vec<u8>` where this is possible, it will remove the restriction of forcing the user to allocate on the heap. This changes happen almost always when a key is asked.

This is a breaking change, we must bump the minor version.

There is another way to make this improvement: Making the methods asking for a key that implement `AsRef<[u8]>`. This way it will not be a breaking change. But I think we didn't make sled reach 1.0 for the moment so we can break things :)

### DIFF
--- a/crates/sled/src/lib.rs
+++ b/crates/sled/src/lib.rs
@@ -5,11 +5,11 @@
 //! ```
 //! let t = sled::Tree::start_default("my_db").unwrap();
 //!
-//! t.set(b"yo!".to_vec(), b"v1".to_vec());
+//! t.set(b"yo!", b"v1".to_vec());
 //! assert!(t.get(b"yo!").unwrap().unwrap() == &*b"v1".to_vec());
 //!
 //! t.cas(
-//!     b"yo!".to_vec(),       // key
+//!     b"yo!",                // key
 //!     Some(b"v1"),           // old value, None for not present
 //!     Some(b"v2".to_vec()),  // new value, None for delete
 //! ).unwrap();
--- a/crates/sled/src/tree.rs
+++ b/crates/sled/src/tree.rs
@@ -154,9 +154,9 @@ impl Tree {
     }
 
     /// Retrieve a value from the `Tree` if it exists.
-    pub fn get(
+    pub fn get<K: AsRef<[u8]>>(
         &self,
-        key: &[u8],
+        key: K,
     ) -> Result<Option<PinnedValue>, ()> {
         let _measure = Measure::new(&M.tree_get);
 
@@ -165,7 +165,7 @@ impl Tree {
         // the double guard is a hack that maintains
         // correctness of the ret value
         let double_guard = guard.clone();
-        let (_, ret) = self.get_internal(key, &guard)?;
+        let (_, ret) = self.get_internal(key.as_ref(), &guard)?;
 
         guard.flush();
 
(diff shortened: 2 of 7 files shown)


crates/sled/src/tree.rs
```diff
@@ -186,22 +186,22 @@ impl Tree {
     /// let t = sled::Tree::start(config).unwrap();
     ///
     /// // unique creation
-    /// assert_eq!(t.cas(vec![1], None, Some(vec![1])), Ok(()));
-    /// // assert_eq!(t.cas(vec![1], None, Some(vec![1])), Err(Error::CasFailed(Some(vec![1]))));
+    /// assert_eq!(t.cas(&[1], None as Option<&[u8]>, Some(vec![1])), Ok(()));
```
Kerollmops: This is a breaking change, the `cas` method now accept an `Option<AsRef<[u8]>>` and this can be a problem for the inference system. If you call `cas` with a `None` in place of the `old` argument you must specify which type is the `Option`.

A solution could be to continue accepting an `Option<&[u8]>`.

crates/sled/src/lib.rs
```diff
@@ -5,11 +5,11 @@
 //! ```
 //! let t = sled::Tree::start_default("my_db").unwrap();
 //!
-//! t.set(b"yo!".to_vec(), b"v1".to_vec());
+//! t.set(b"yo!", b"v1".to_vec());
 //! assert!(t.get(b"yo!").unwrap().unwrap() == &*b"v1".to_vec());
 //!
 //! t.cas(
-//!     b"yo!".to_vec(),       // key
+//!     b"yo!",                // key
```
spacejam: I'm unreasonably excited about being able to write simple demos without having to litter .to_vec() etc... everywhere lol :)

crates/sled/src/tree.rs
spacejam: Grrr https://github.com/rust-lang/rust/issues/20063 strikes again... In this case, I'd prefer to keep the signature for cas the way it was. The old value is flexible when it stays as an `Option<&[u8]>`, and we need to allocate anyway to fill the new value if it's `Some`. The clone that we do while creating the node can be optimized away by using an `Option` above the loop that is set before the loop with the provided value, taken to fill the new node, and if the inner CAS fails we can take the value from the created node and reuse it in the next loop.

No need to implement the zero-additional-copy `Option` trick, but I think it's better to keep the API as it is for `cas` because it communicates the need to allocate. The other changed methods need to allocate too, and they should also implement this trick to remove their inner `clone` calls, but I'm OK with them accepting any `AsRef<[u8]>` because it makes trying out the system much easier. 

It's a bit unidiomatic to not require an owned value to be passed in, but in this case it does let us simplify usage in some cases, especially for new people who are just trying it out, and I want to make this easier where possible. It's at conflict with the `cas` method staying explicit and owned, but I'm OK with that in this case.

Kerollmops: I just though about `Into<Vec<u8>>` combined with `AsRef<[u8]>`, but it will complexify the API a little bit. Using these bounds we could avoid duplications if the user gives a `Vec<[u8]>` or a `Box<[u8]>` as key but will allocate if it's a raw `&[u8]` for example.

In this case doesn't a combination of `Borrow` and `Eq` better fitting ?
Like the [`HashMap::get`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get) method.

I think we should just make this PR effective now and think about that (or not) in another place.

Kerollmops:
I updated the `cas` method to accept an `Option<&[u8]>` instead of an `Option<AsRef<[u8]>>`.

spacejam:
looks good, thanks!!!

PR #433, opened 2018-11-16
Allow path to be overridden for temporary databases. Closes #431.

### LINKED ISSUE #431: Surprised by behavior of "temporary"
I built a sled instance with `.temporary(true)`, and was surprised that as a result it chose to place my database/pagecache in `/dev/shm` regardless of the path I gave it.

My `/dev/shm` is a ~4GB tmpfs and quickly ran out of space, resulting in an OOM, followed by all sorts of quirky behavior on my machine (e.g., chrome inexplicably crashing). I was just expecting my designated path to be removed when the `Config` was dropped, but looking at the code this is clearly not the intent of that setting.

Initial thoughts...
1. I think it'd be nice if `temporary` were documented better -- I.e., that it stores the database in memory (/dev/shm) and your path is ignored.
2. Maybe a new function ("auto_remove"?) would also be nice for my use-case (I'd like to use memory but don't have enough room in memory so I'm using disk instead), so I don't need to cleanup manually.

spacejam:
cc @djahandarie

### DIFF
--- a/crates/pagecache/src/config.rs
+++ b/crates/pagecache/src/config.rs
@@ -19,6 +19,8 @@ use serde::Serialize;
 
 use super::*;
 
+const DEFAULT_PATH: &'static str = "default.sled";
+
 impl Deref for Config {
     type Target = ConfigBuilder;
     fn deref(&self) -> &Self::Target {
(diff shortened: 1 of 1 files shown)


djahandarie:
LGTM! Thank you!!

spacejam:
looking more deeply into the dropping issue as well, it's been bugging me for a while

Issue #435, opened 2018-11-17
flatbuffer representation of Tree nodes


spacejam:
A large amount of time is now being spent in bincode serializing and deserializing. It's gotten us pretty far, but now it's time to optimize these costs way down by using flatbuffers. This will also move us toward the 1.0 goal of having a forward-compatible binary format.

ozgurakkurt:
could this be done with capnp istead of flatbuffers since it's rust support is better

spacejam:
@ozgurakkurt it could be done with a wide variety of options, but flatbuffers is among the most human-friendly options that supports first-class schema migration. I see it as the least risky for a number of reasons.

spacejam:
closing this out for now, as at this point it's most likely that a custom parser will be written, as we have been locking dependencies down pretty strongly and we have pretty low churn in Node representations at this point

PR #451, opened 2018-11-27
WIP: flatbuffer representation of Tree nodes #435

utaal:
I'm giving #435 a try.

Currently, I've only removed the `(De)Serialize` bounds, and the `(de)serialize` calls, I'm now trying to figure out FlatBuffers.

### DIFF
--- a/crates/pagecache/src/config.rs
+++ b/crates/pagecache/src/config.rs
@@ -12,10 +12,10 @@ use std::{
     },
 };
 
-use bincode::{deserialize, serialize};
+// use bincode::{deserialize, serialize};
 use fs2::FileExt;
-use serde::de::DeserializeOwned;
-use serde::Serialize;
+// use serde::de::DeserializeOwned;
+// use serde::Serialize;
 
 use super::*;
 
@@ -47,7 +47,7 @@ impl Deref for Config {
 ///     .path("/path/to/data".to_owned())
 ///     .read_only(true);
 /// ```
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Clone)] // TODO(flatbuffers)
 pub struct ConfigBuilder {
     #[doc(hidden)]
     pub blink_node_split_size: usize,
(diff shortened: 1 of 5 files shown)


spacejam:
Hey! I've written up a flatbuffer schema and generated some code for it here: https://github.com/spacejam/sled/tree/tyler_flatbuffers

Feel free to use it or not! But this may minimize the conversion effort. Let me know if I can help at all!

jeehoonkang:
I'm interested in picking it up from here.  @spacejam would you please provide me with some guidance?  More concretely, what should I do for you to merge this PR?

spacejam:
Hey @jeehoonkang! The first thing to know is that this is pretty experimental work, and we don't know what kind of impact this will have on performance yet, and it may very well slow things down.

That said, I imagine recovering items as defined in the above flatbuffer specification and maybe wrapping them in something like this, as if it were applied to the flatbuffer example at https://github.com/google/flatbuffers/tree/master/tests/rust_usage_test:
 
```
struct MonsterBox {
    _buf: Box<[u8]>,
    mon: Monster<'static>,
}

impl MonsterBox {
    fn from_buf(buf: Vec<u8>) -> MonsterBox {
        let mut buf: Box<[u8]> = buf.into_boxed_slice();
        let buf_ptr = buf.as_mut_ptr();
        let static_slice = unsafe {
            std::slice::from_raw_parts(buf_ptr, buf.len())
        };
        let mon = my_game::example::get_root_as_monster(
            static_slice,
        );

        MonsterBox { _buf: buf, mon }
    }
}

impl std::ops::Deref for MonsterBox {
    type Target = Monster<'static>;

    fn deref(&self) -> &Monster<'static> {
        &self.mon
    }
}
```

There remains an open question about how to integrate the IVec type that currently uses an Arc to back keys and values in the system, as they can really help to keep some in-memory fragment chain compaction stuff fast and low memory pressure. It would also be fine to go back to the old way of protecting buffers to users coupled to crossbeam `Guard`'s to facilitate zero-copy reads instead of the `IVec`-style `Arc`. If this is something you would be interested in playing with and trying to get a nice story for, it would be awesome, but just know that this is all kind of high-risk work in that we don't know what will actually result in better performance or not, and we may end up implementing it only to throw it away after terrible measurements.

spacejam:
Closing this due to inactivity, but over time I'm interested in returning to this

PR #455, opened 2018-12-01
sled: use saturating u64 for size calculations

lucab:
This introduces u64 and saturating adds for size-related math internal
to sled.


spacejam:
looks good, thanks @lucab  :]

PR #456, opened 2018-12-05
Optionally measure allocations

spacejam:
@utaal Just added a pretty simple allocation counter based on @rw's counter in the flatbuffers codebase, but using AtomicUsize so it works across threads.

recommended to run from benchmarks/stress2:
```
 cargo run --features=measure_allocs --  --threads=1
```

non-release mode on my laptop does 250k high-level operations in 10 seconds, generating nearly 4 million allocations with about 840mb of total allocation.

release mode: roughly 10x higher in all measurements over 10 seconds.

TONS of low-hanging fruit :] Now this can be a fun game to get the number down


davidrusu:
very cool!

PR #458, opened 2018-12-11
Sweep over things, adapt to 2018

spacejam:


spacejam:
pretty boring migration since we already enabled the 2018 lints a few months back

PR #470, opened 2018-12-18
windows and clippy

siler:
Hey, here are a bunch of changes to look at. Most were driven by wanting to compile/pass tests on windows and fixing clippy warnings. This project is really exciting.

### DIFF
--- a/benchmarks/stress2/src/main.rs
+++ b/benchmarks/stress2/src/main.rs
@@ -1,3 +1,9 @@
+extern crate docopt;
+extern crate rand;
+extern crate serde;
+
+use serde::Deserialize;
+
 use std::{
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
--- a/benchmarks/stress2/src/main.rs
+++ b/benchmarks/stress2/src/main.rs
@@ -1,10 +1,3 @@
-#[macro_use]
-extern crate serde_derive;
-extern crate docopt;
-extern crate env_logger;
-extern crate rand;
-extern crate sled;
-
 use std::{
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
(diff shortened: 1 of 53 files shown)


siler:
Any ideas on the fuzz test? I'm still looking. Not feeling brave enough to take on reserve_inner today.

spacejam:
this is a bit much to review, would you mind splitting this into smaller PRs? this will make it more clear whatever is breaking the osx build etc...

siler:
Absolutely, I'll split it up tonight.

A couple of related questions:

#393 says that you'd rather not change the tree interface to use slices at the moment, is that the case for any other APIs? Also, is it acceptable to tag them with clippy allows for now?

Also, there are some places where a deref is used where it isn't strictly necessary. Usually it is around k/v handling. I'm curious if that is to be more explicit about the behaviors that will be taken or if it is just because it works. An example is using &* on a Vec<u8> -> &[u8] conversion instead of just &/as_slice. I think in a number of cases the deref can be safely removed (something like 30+). Thoughts?

spacejam:
* In general, prefer not adding clippy exceptions, and keeping it breaking overall until higher level concerns are addressed
* The API should generally match BTreeMap wherever possible to minimize surprises. There are a few places where we clone keys and values that are passed in by value, but these will be optimized away and passing them in by value is desired, despite it currently not actually being necessary due to the inner clones. 
* The `AsRef` instances should actually be `Borrow` instead, since `Borrow` communicates that the passed-in parameter retains `Ord`, which is important. `AsRef` may break the `Ord` relation. This should be its own PR, and I may just do it right now actually since it's low effort

spacejam:
hmm actually I'm OK with keeping it as AsRef for now, since it allows us to have a more accepting API and the number of things that break the Ord guarantee for when they are specifically AsRef<[u8]> is low AFAIK.

PR #473, opened 2018-12-23
more clippy!

siler:
Here's another bunch to look at. Only bigger ones are the match dereference simplifications and the option macro changes. Clippy was complaining about the macro's 'check' argument being immediately operated on by !, which seems reasonable.


siler:
After these the only ones left are a bunch of needless_pass_by_value and one cyclomatic complexity. :)

crates/pagecache/src/config.rs
```diff
@@ -515,43 +516,46 @@ impl Config {
...
-            self.inner.compression_factor <= 22,
+        check_supported!(
+            self.inner.compression_factor > 22,
             "compression_factor must be <= 22"
         );
-        supported!(
-            self.inner.idgen_persist_interval > 0,
-            "idgen_persist_interval must be above 0"
+        check_supported!(
+            self.inner.idgen_persist_interval < 1,
```
spacejam: <=

spacejam: whoops, nvm

spacejam: is "supported" a clippy warning? I think I would prefer to keep the conditionals the way they were

spacejam:
woo so nice :)

Other than the one nit about conditionals in the `supported` macros this looks good to me. Thanks so much for cleaning all of this up :]

crates/pagecache/src/config.rs
siler: Not supported specifically, I can change the word back. Here's the clippy error: https://rust-lang.github.io/rust-clippy/master/index.html#neg_cmp_op_on_partial_ord. So to solve it I inverted the condition. This way it also follows flow similar to an if statement, which is why I added the word "check" to reinforce that it is a condition. I can change the name back if you'd like. Any thoughts on avoiding inverting the condition, if that's an issue?

spacejam: I'd prefer to keep it more like an assert, where the condition must be true, and supported to stay the same. For that specific error, feel free to throw an outer `!` around the PartialOrd positive comparison

spacejam:
Looks great, thanks again for cleaning all this stuff up! I learned a few things about attributes from your changes, too :D

PR #478, opened 2018-12-30
keyspace support!!!

spacejam:


Kerollmops:
Awesome !
Are keyspace some kind of RocksDB column families ?
According to [the RocksDB terminology](https://github.com/facebook/rocksdb/wiki/Terminology) it seems like it is 😃

spacejam:
@Kerollmops yeah! The goal is to share multiple Trees (and eventually other kinds of structures) in the same pagecache, ID generator and transaction tracker so that transactions can be performed across multiple structures in different tables / keyspaces / column families over time!

Kerollmops:
Do you think that the `Db` must always implement the `Deref` trait ?
Now that the `Db` can open any `Tree` is it always correct to deref to the "default" `Tree` ?

PR #493, opened 2019-01-09
Clean the Db Deref implementation

Kerollmops:
It seems weird to me, I think it's probably related to a piece of old code.
I am probably wrong :)


spacejam:
Ahh much better, thanks for cleaning this up :D

PR #495, opened 2019-01-09
skip mark link on previously linked

spacejam:
this will significantly reduce contention on the segment accountant mutex


spacejam:
This reduces mutex acquisitions for the segment accountant by 90% on the stress2 benchmark with 10000 total keys across 8mb segments

Issue #492, opened 2019-01-09
Opening sled Db with a different Config for each Keyspace


Kerollmops:
Currently the `Db` can be opened using a `Config` struct, this configuration is the same for all keyspaces. This can be a problem when a keyspace specific merge operator must be used for example.

I know that an API clean up will be done at a moment, I just want to know if it is possible technically to do so.

In the RocksDB universe, the `DB` must be opened by specifying the name/config of each column family you want to open. It seems that if you don't want to open them all for some reason, it is an error. With sled, all of them are opened automatically but we cannot define keyspace specific settings.

Here is a possible API and an example usage that does not add too much complexity.

```rust
fn Db::with_tree_config(path: P, config: Config, tree_config: F) -> Result<Db, ()>
where
    P: AsRef<Path>,
    F: Fn(&[u8], ConfigBuilder) -> Config,
{
    // this is the base settings which will be passed to the tree config generator
    // deref gives us a ConfigBuilder
    let builder = (*config).clone();

    for name in tree_names {
        let config = (tree_config_generator)(name, builder.clone());
        // ...
    }

    // ...
}
```

```rust
fn config_by_name(tree_name: &[u8], builder: ConfigBuilder) -> Config {
    let builder = match tree_name {
        b"first-tree"  => builder.setting1(3),
        b"second-tree" => builder.destroy_the_world(true),
        b"third-tree"  => builder.merge_operator(the_lol).setting1(4),
        _              => builder,
    };

    builder.build()
}

let config = ConfigBuilder::new().build();
let db = Db::with_tree_config("sled/db/path.sdb", config, config_by_name)?;
```

Something that is not pleasant with this API is that the `Config`/`ConfigBuilder` types should be replaced by `TreeConfig`/`TreeConfigBuilder` which will be more specific to `Tree`s.

The `ConfigBuilder` could generate a `TreeConfigBuilder` that does expose only the rational settings.

I am not entirely convinced by this API, there should probably be something more Rust idiomatic.

spacejam:
Hey! This is also something I've been thinking about, and it's definitely possible to set per-`Tree` merge operators. At this point, it's just about figuring out the right API for making it as simple as possible.

I'd like to make `Config` less complicated, as the implementation is already kind of a ratnest of complexity around file handling and thread pools etc... 

[...]

PR #509, opened 2019-01-27
Improve the ways of opening Db and Trees

Kerollmops:
Here is some work on fixing #492, I thought of a builder constructor that follows the [std::fs::OpenOptions](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html) type.

I would be cool to guide me a little bit about the `Context` type you talked about, it will help me continue working on the `OpenOptions` types. What do we put in it? Where does it lives, pagecache or sled?

Here is the list of things I need to do:
 - [ ] Introduce some `OpenOptions`/`TreeOpenOptions` type constructors.
  The `OpenOptions::merge_operator` setting will be passed to the `TreeOpenOptions::merge_operator` to open the default tree.

- [ ] Remove the `Db::start_default/start` methods.

 - [ ] Remove the `pagecache::Config::merge_operator` setting, it does not need to care about, only the `Tree`s must. So, one serialized config by tree?

 - [ ] Introduce a `Context` type that owns a `PageCache`.
  This type is aware of the settings relative to a `Tree`, it must be serializable, this way a `Tree` can be loaded again with the same settings (e.g. the merge operator) and reduce the introduction of bugs related to settings.

 - [ ] Remove the `Config::temporary` and the `DEFAULT_PATH` setting, it could be better to make `Config`s parametric on `AsRef<Path>` and being able to use [`TempDir`](https://docs.rs/tempfile/3.0.5/tempfile/struct.TempDir.html).

### DIFF
--- a/crates/sled/Cargo.toml
+++ b/crates/sled/Cargo.toml
@@ -33,5 +33,5 @@ version = "0.4"
 
 [dependencies]
 pagecache = { path = "../pagecache", version = "0.9.4" }
-sled_sync = { version = "0.2.2", path = "../sled_sync" }
+sled_sync = { path = "../sled_sync", version = "0.2.2" }
 futures = "0.1"
--- a/crates/sled/src/db.rs
+++ b/crates/sled/src/db.rs
@@ -1,5 +1,6 @@
 use std::{
     ops::Deref,
+    path::Path,
     sync::{
         atomic::{
             AtomicUsize,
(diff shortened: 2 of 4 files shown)


spacejam:
Hey! This is something that I'd like to work on after the work on transactions is a bit more fleshed out. The reason is that by adding this complexity into an implementation right now, it makes the effort required to define a generic transactional layer across multiple collection types more constrained. I'd like to avoid solving this problem for now, to try to keep transaction implementation friction lower. I'm pushing back on this because that friction is already extremely high, and a number of interfaces may need to be redesigned. 

Per-tree configurability is not hard in theory, and I don't expect this to be problematic once unified transactions are implemented, but I'd like to keep things like this completely out of the way for now to reduce the effort going into that.

I appreciate the effort you're putting into this, but in the future it might be helpful to get buy-in on configuration changes in issues before beginning implementation. I'm not opposed to this work, but I'd like to keep it out of the way until transactions are in a better specified and implemented place.

PR #546, opened 2019-02-18
Add concurrent log test

### LINKED ISSUE #404: add concurrent log test
several threads should share a log,
* read current counter
* reserve log buffer with read counter + 1
* cas counter old -> old + 1
* abort or complete res based on cas
* if successful, success counter += 1
after all threads do this a few million (?) times, join them and drop the log
ensure we're able to read from 0 to success counter sequentially

anxiousmodernman:
Share a log among several threads, and also share an AtomicUint counter.

Threads will attempt to successfully CAS on the counter to increment it by 1. 
If this succeeds, that signals to the test complete its log reservation,
otherwise it will abort.

This counter indicates the number of sucessful writes to the shared log.
The contents of the write will be the value read from the counter.

Finally, iterate through the log, and assert that the number of
successful entries exactly matches the success counter, and that the
values increment sequentially.

Refs #404

### DIFF
--- a/tests/tests/test_log.rs
+++ b/tests/tests/test_log.rs
@@ -18,6 +18,7 @@ use {
         thread,
     },
 };
+use std::mem::size_of;
 
 type Lsn = i64;
 type LogId = u64;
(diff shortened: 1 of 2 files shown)


anxiousmodernman:
100k iterations runs in ~3 min on my new macbook. I cut it to 50k, but since it's an integration test we might be comfortable with more.

I did not delete the existing concurrent log test. We could do that, as well, or leave it.

tests/tests/test_log.rs
```diff
@@ -161,6 +162,62 @@ fn concurrent_logging() {
...
+    }
+
+    drop(log_arc);
+
+    let log = Log::start_raw_log(config).unwrap();
+    let mut iter = log.iter_from(SEG_HEADER_LEN as Lsn);
+
+    for i in 0..SUCCESS_COUNTER.load(Ordering::SeqCst) {
+        // Do we care what's in the log? or just that we can iterate?
+        let (_, _, _) = iter.next().expect("expected some");
```
spacejam: as the final bit of correctness checking, we should check here that the read value matches `i`, since this test will serve as a validation of the core ordering assumption in `iobuf.rs`

tests/tests/test_log.rs
```diff
@@ -161,6 +162,62 @@ fn concurrent_logging() {
...
+                let iterations = 50_000;
+                for i in 0..iterations {
+                    let current = SHARED_COUNTER.load(Ordering::SeqCst);
+                    let raw_value: [u8; size_of::<usize>()] = unsafe { std::mem::transmute(current+1)};
+                    let res = log.reserve(raw_value.to_vec()).unwrap();
+                    match SHARED_COUNTER.compare_and_swap(current, current+1, Ordering::SeqCst) {
+                        // If the current value was returned, then CAS succeeded on AtomicUsize
+                        current => {
+                            res.complete();
+                            SUCCESS_COUNTER.fetch_add(1, Ordering::SeqCst);
```
spacejam: we can actually totally rely on `SHARED_COUNTER` here, since it will equal `SUCCESS_COUNTER`, so we can remove the second counter and just load `SHARED_COUNTER` below when iterating / checking

tests/tests/test_log.rs
```diff
@@ -161,6 +162,62 @@ fn concurrent_logging() {
...
+    static SUCCESS_COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+    let num_threads = 6;
+    let mut handles = Vec::new();
+    for t in 0..num_threads {
+        let log = log_arc.clone();
+        let h = thread::Builder::new()
+            .name(format!("t_{}", t))
+            .spawn(move || {
+                let iterations = 50_000;
```
spacejam: This is probably good to set to a const at the top of the function, maybe set to something more like 5000 to shave off a few minutes from the test run but still making it pretty easy to bump up in targeted testing

spacejam:
Just a couple minor points, but overall this looks great! Thanks for helping out with this :D

anxiousmodernman:
Removing one of the atomic uint counters, at least the way I've done it, reliably panics this test when run multi-threaded. 🤔 

Maybe I'm doing it wrong. [Should this work?](https://github.com/spacejam/sled/pull/546/files#diff-41ba2d63ebc862a9fc5b74dd830df613R189) Or is there too much of a race between a thread's successful CAS and the reservation complete? It seems like we should be able to rely on successful CAS operations as our "success" counter, because only a successful CAS will take the codepath to call `reservation.complete`.

For comparison, [here is what worked before with 2 counters, forked to a different branch](https://github.com/anxiousmodernman/sled/blob/404_original_concurrent/tests/tests/test_log.rs#L175-L219).

spacejam:
I had to play with this for a really long time before the bug jumped out... ugh...

the first match arm will never fail. `current` should be `c if c == current`.

here's some code for verifying that it all zips together in the for loop below
```
    for i in 0..successfuls {
        let val = iter.next().expect("expected some log entry");
        let mut raw_value = [0_u8; size_of::<usize>()];
        raw_value.copy_from_slice(&*val.2);

        let read: usize = unsafe { std::mem::transmute(raw_value) };
        assert_eq!(
            i + 1,
            read,
            "failed to read expected value at log coordinate {}",
            val.1
        );
    }
```

PR #549, opened 2019-02-20
Use a cached AtomicUsize tree root so it can fast-path around the pag…

spacejam:
…ecache to read the root each time


spacejam:
superceded by https://github.com/spacejam/sled/pull/554

PR #550, opened 2019-02-22
update flamegraph dependency

spacejam:


ehiggs:
Wouldn't this be better as a straight submodule of https://github.com/brendangregg/FlameGraph?

spacejam:
@ehiggs if the dependency changed frequently and we needed to update frequently then maybe, but neither is true, and the overhead for me of entering git submodule commands would be higher than the effort that this took

PR #551, opened 2019-02-22
Reduce allocations in the log reservation path

spacejam:


tests/tests/test_log.rs
```diff
@@ -274,34 +286,33 @@ fn log_iterator() {
...
     {
-        let res =
-            log.reserve(b"never_gonna_hit_disk".to_vec()).unwrap();
+        let res = log.reserve(b"never_gonna_hit_disk").unwrap();
         res.abort().unwrap();
     }
 
-    log.write(b"4444".to_vec());
-    let (last_lsn, _) = log.write(b"55555".to_vec()).unwrap();
+    log.write(b"4444");
```
Kerollmops: Just for information: [the primitive `str` type implement `AsRef<[u8]>`](https://doc.rust-lang.org/std/primitive.str.html#impl-AsRef%3C%5Bu8%5D%3E), it means that it is possible to directly call `log.write("4444")`.

PR #564, opened 2019-02-26
Nuke PinnedValue. Use Tx structure in PageCache instead of raw crossb…

spacejam:
…eam Guard


spacejam:
@Kerollmops yeah, we want to basically inline 7 bytes on 32-bit machines and 15 bytes on 64-bit machines. A slice is 2 pointer widths, since it has a pointer and a usize (which defaults to the pointer size, which might be different from the machine's word size)

PR #570, opened 2019-03-05
Several orthogonal improvements

jeehoonkang:
Here are some orthogonal improvements:

- Rust code highlight in pagecache's README
- Use `guard.defer_destroy()` instead of `guard.defer()`.  See [the API doc](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-epoch/src/guard.rs#L199).
- Create `Tx::new()`.
- Fix a documentation bug.
- Use `Relaxed` instead of `SeqCst` in pagetable's `drop`.  It is sound because the destructor is accessing the data without contention.


spacejam:
Hey @jeehoonkang, nice to see you around here again :D Thanks for these cleanups! Will merge as soon as the tests go green

PR #585, opened 2019-03-17
Make the IVec Debug impl support formatting

Kerollmops:
It is now possible to specify formatting rules, the "underlying" slice will be informed of the outer formatting rules. For example the `{:x?}` will now correctly display the slice in its hex representation.


spacejam:
Thanks!

PR #586, opened 2019-03-17
[wip] restructure results. closes #532. closes #525. closes #531. closes #583.

### LINKED ISSUE #525: remove danger_cast
this is a common source of panics in highly concurrent code, most recently:
```
   6: sled::tree::Tree::recursive_split
             at /home/t/src/sled/<::std::macros::panic macros>:3
             at crates/sled/src/tree.rs:1113
             at /rustc/9fda7c2237db910e41d6a712e9a2139b352e558b/src/libcore/result.rs:530
             at crates/sled/src/tree.rs:1111
             at crates/sled/src/tree.rs:945
```

it's time to completely rethink how CAS failures are propagated internally and externally

### LINKED ISSUE #531: have pagecache return failed writes to avoid copies in cas loops in tree

### LINKED ISSUE #532: split CasFailed from Result
IO errors should be handled in a totally different context from CAS failures, so it only makes code more bug-prone to mix the results.

A large number of race conditions have been found relating to the mishandling of these two separate concerns, so we should completely separate these at the type level instead of bug-prone runtime assertions.

### LINKED ISSUE #583: maintain 69 issues at all costs
nice

spacejam:


.rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+reorder_imports = true
+max_width = 60
```
Kerollmops: Why do you specify a `max_width` of 60 ? I had this feeling that it is less possible to write function calls on one line and make things a little less easy to read.

spacejam: good catch, I just copied it from a project and also noticed I didn't like how it was reformatting things

Kerollmops:
Wow great work !

PR #588, opened 2019-03-18
Implement AsRef for IVec

Kerollmops:
Implementing this trait make the `IVec` more easy to use with comman type (i.e. [`io::Cursor`](https://doc.rust-lang.org/std/io/struct.Cursor.html)).

Do you think the `IVec` can be considered a smart pointer? https://rust-lang-nursery.github.io/api-guidelines/predictability.html#c-deref


spacejam:
Looks good! More functionality in a net-red diff is always nice :D

Re: smart pointer, I believe so, because this is essentially a read-only Arc<[u8]> with an inline implementation for small values

PR #590, opened 2019-03-21
Reduce the number of times a value is cloned

Kerollmops:
This pull request reduce the number of times a value is cloned in the `set`, `merge` and `cas` Db methods. It can be costly especially for big values.

To fix that I directly put the user value inside of an `IVec` this way cloning is not costly. By the way I changed the `Event::Set/Merge` types to be more consistent with the current Db API, returning the values as `IVec` and not `Vec<u8>`.

### DIFF
--- a/crates/sled/src/subscription.rs
+++ b/crates/sled/src/subscription.rs
@@ -15,15 +15,17 @@ use futures::{
     },
 };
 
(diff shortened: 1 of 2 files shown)


spacejam:
Looks great, thanks @Kerollmops :D

PR #591, opened 2019-03-22
Check if the compression feature is enabled

Kerollmops:
This pull request change two things, it enables the pagecache compression feature checks if the compression feature has been set when `use_compression` is set to `true`.

I faced the problem of running sled without the `compression` feature active but with the `use_compression` config builder param set to `true` (the default value) and sled didn't tell me anything.

Even worst: I created a database with the `compression` feature and tried to read it, get values from trees, with another program without the `compression` feature active, sled did not say anything but did not returned any entry, like they were empty.

So there is two solutions to fix this problem:
  - keeping the `use_compression` param set to `true` by default **but** make the `compression` feature enabled by default or
  - set the `use_compression` to `false` by default and keep the `compression` feature disabled

In this PR, the first solution has been chosen.

EDIT: The second solution is preferable.

### DIFF
--- a/crates/pagecache/src/config.rs
+++ b/crates/pagecache/src/config.rs
@@ -34,7 +34,6 @@ impl Deref for Config {
 /// let _config = pagecache::ConfigBuilder::default()
 ///     .path("/path/to/data".to_owned())
 ///     .cache_capacity(10_000_000_000)
-///     .use_compression(true)
 ///     .flush_every_ms(Some(1000))
 ///     .snapshot_after_ops(100_000);
 /// ```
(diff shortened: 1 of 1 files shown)


crates/pagecache/Cargo.toml
```diff
@@ -12,7 +12,7 @@ readme = "README.md"
 edition = "2018"
 
 [features]
-default = []
+default = ["compression"]
```
Kerollmops: Do you think it could be better to enable it in the sled toml too? Even if it is already a default feature of pagecache? For clarity?

spacejam:
Hey! Yeah, this was a little hack to just get compression to be enabled when people flip on the flag, but this is a nasty issue. I'd like to minimize default dependencies, which will keep CI running faster also etc... So I think I would prefer option 2, and keeping your supported check.

Kerollmops:
Ok great! I will change that right away then :)

Kerollmops:
It strange that I get some doc tests not passing on my mac. (the `src/lib.rs` -  (line 5) and the `src/tree.rs` - tree::Tree (line 31)).
With the same error: `index out of bounds: the len is 0 but the index is 0` resulting for an `Vec::<T>::IndexMut` inside of the `pagecache::segment::SegmentAccountant::mark_replace` function.

PR #592, opened 2019-03-23
Prefer inserting at index instead of sorting

Kerollmops:
The `get_internal` operation moved from 25% to 15% and make my program take from 28mins to 25mins.

[flamegraphs-before-after-patch.zip](https://github.com/spacejam/sled/files/2999208/before-after-patch.zip)


spacejam:
Very nice!!! Thanks :D

PR #593, opened 2019-03-23
Rewrite the Data split method

Kerollmops:
I tested and assumed that the `ptrs` are already sorted and that prefix encoded keys are sorted the same that non encoded ones (is that named "ordering transitivity" ?).

I will rebase on master once #592 is merged, if it is 😄 

So in this rewrite I no longer need to construct a decoded sorted Vec but only need to decode the middle key (`split`) and create a new Vec of half the size of `ptrs` re-encoded using the previously extracted middle key (`split`).

I also rewrite the `prefix_decode` and `prefix_encode` functions removing the unsafe code that doesn't seems to be needed and introduced a new `prefix_reencode` function that `decode` and `encode` using the the old and new prefix.

This update reduced the usage percentage of the `Data::split` method from 46.2% to 25.47% and make my program takes from 25mins (including the #592 patch) to near 16mins.

[flamegraphs-before-after-data-split-patch.zip](https://github.com/spacejam/sled/files/2999752/before-after-data-split-patch.zip)

### DIFF
--- a/crates/sled/src/data.rs
+++ b/crates/sled/src/data.rs
@@ -1,4 +1,3 @@
-use std::fmt::Debug;
 use std::mem::size_of;
 
 use super::*;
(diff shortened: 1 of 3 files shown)


spacejam:
Beautiful! So excited to run my benchmarks on my new dev machine with your changes!

Kerollmops:
When will you do that? I have many other improvements to do 😄

Kerollmops:
No, no don't merge master, I am rebasing!

spacejam:
Possibly a bit later today, but for now I need to run outside as it's the first time you can wear a short sleeved shirt in the sun outside today in Berlin :D I'll merge this as soon as tests go green!

spacejam:
whoops, absent minded staring out the window, apologies!

Kerollmops:
Same good weather in Paris yesterday! Have outside outside, away from internet and code 👍

Kerollmops:
Do not merge that please, I am currently working on another commit that I would want to put into this PR.

spacejam:
This is really good stuff, I'm going to run to my office in a little bit to do some simple benchmarks and I'm really looking forward to running this! 

Great job!!!

Kerollmops:
By the way is the prefix encoding done to reduce the key sizes?

spacejam:
Yeah! On a lot of datasets where most of the keys on a single page will be
the same except the last byte, it basically means we get free key storage

Kerollmops:
So in MeiliDB, when I create 40 keys by document that all starts with the _document id_ and are followed by an incrementing number (from 0 to 39 representing each field), there is only two bits stored for each documents + 1x the prefix?
So cool !

spacejam:
yeah :D at some point I want to look into using SIMD to speed up the comparison even more, because every operation needs to do this several times to get to the leaf node they're interested in

Kerollmops:
Yes, I though about that too when rewriting these and searched for "longest prefix algorithms" but found nothing interesting (all using Binary Trees or likely data structures). It will probably be an hand-crafted algortihm... :)

Kerollmops:
And here we are, I do not tried to put it into sled again, but you can see a not quite advanced version of a longest-common-prefix function using some simd (the current version is limited to `x86-64` `AVX2` only).

https://github.com/Kerollmops/longest-common-prefix

PR #597, opened 2019-03-29
some write optimizations

spacejam:
@Kerollmops for workloads I tried with larger values (800 bytes+), this caused a really major speedup. Basically, we were doing a ton of wasted work trying to split nodes even when we detected that we were failing some of the recursive tree splits. Now we bail as soon as we fail once, because it's incredibly likely that we will not win any others either.

### DIFF
--- a/crates/pagecache/src/logger.rs
+++ b/crates/pagecache/src/logger.rs
@@ -437,7 +437,6 @@ impl Log {
         let mut header = iobuf.get_header();
 
         // Decrement writer count, retrying until successful.
-        let backoff = Backoff::new();
         loop {
             let new_hv = iobuf::decr_writers(header);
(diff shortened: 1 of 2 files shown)


Kerollmops:
Hi @spacejam! I tested this update on my own data and sadly it doesn't improve significantly the time spent child split. Here is a flamegraph.

[flamegraph-create-database.svg.zip](https://github.com/spacejam/sled/files/3024264/flamegraph-create-database.svg.zip)

spacejam:
I've been going down a bit of a rabbit hole and realized that there are
some big issues with root splitting, which might also explain some of the
space issues you were encountering with space amplification

Kerollmops:
Great to here that the probable reason of the issue are some kind of identified. I will never be enough thankful to the job and research you do!

PR #601, opened 2019-03-31
debug gc

spacejam:


Kerollmops:
Hi Tyler! Here are some of the stats I get with my workload! You have done a great job, that's absolutely wonderful 🎉

## Using these custom settings

```rust
.io_bufs(16)
.io_buf_size(8_000_000)
.async_io_threads(3)
.blink_node_split_size(4096)
.page_consolidation_threshold(10)
.cache_bits(0)
.cache_capacity(1_000_000_000)
.flush_every_ms(Some(200))
.snapshot_after_ops(100_000_000_000)
.print_profile_on_drop(true)
```

### before

create-database without compression initial was 454Mb results in 47Go in 780.54s (13min).

query-database loaded in 415.78s the first time and second time in 1.13s.

### after

create-database without compression initial data was 454MB results in 1.7Go in 83.5s.

query-database loaded in 7.97s the first time and second time in 1.13s.

## Using the default settings

### after

create-database without compression initial data was 454MB results in 1.8Go 82.63s.

query-database loaded in 3.36s the first time and second time in 1.22s.

spacejam:
Wow, that's huge! I'm quite happily surprised you're seeing such a nice improvement on both size and total runtime. And there are still tons of optimizations to apply :)

PR #602, opened 2019-04-01
[merge after april 11 (1.34)] Cleanup of conversions, never downcast without a check or reasonable confidence in correctness, remove nightly feature

spacejam:
@lucab this one's for you :)


crates/pagecache/src/iobuf.rs
```diff
@@ -41,9 +32,9 @@ macro_rules! io_fail {
 
 pub(crate) struct IoBuf {
     pub(crate) buf: UnsafeCell<Vec<u8>>,
-    header: CachePadded<AtomicUsize>,
-    lid: AtomicUsize,
-    lsn: AtomicUsize,
+    header: CachePadded<AtomicU64>,
+    lid: AtomicU64,
+    lsn: AtomicLsn,
```
lucab: This moves from unsigned to signed. Is there a case for LSN to be negative? Can they be `AtomicU64` instead (I think it may make reasoning easier)?

spacejam: Yeah, lsn (logical sequence number) starts at -1 because it's often used through the codebase to represent the current offset that is actually stable on disk. Because the disk offset is a u64, it's also nice as a defensive coding thing to make it a different type, which would have prevented a few stupid bugs in the past where I conflated logical ID for physical location

Kerollmops:
You could probably remove all of those functions too (thanks to the [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) stabilization).
https://github.com/spacejam/sled/blob/master/crates/pagecache/src/util.rs

Issue #587, opened 2019-03-17
CRDT-style page recovery


spacejam:
We can use the log recovery coordinates as a CRDT-style LWW-register for page fragments. This allows us to fully parallelize recovery after crashes, with the only thing that needs to be done serially being the detection of tears in the unstable tip.

Kerollmops:
Ok so, I [read about the CRDT](https://medium.com/@amberovsky/crdt-conflict-free-replicated-data-types-b4bfc8459d26) and searched what does LWW-regsiter means (_last-writer-win_).

Could you please provide some more information of how and where in the code we could work on this? It would be really helpful :)

spacejam:
Part of this will involve parallel snapshot creation, specifically the log recovery iterator should move like this:
* iterate over the dynamic unstable tail on a single thread, in a forward direction
* then iterate over everything else in a reverse direction and in parallel

The reason for reverse iteration when moving beyond the dynamic unstable tail is to reduce the amount of state that is churned out in a shared snapshot. By filling with new information first, we can skip old replaced information immediately without adding it to a shared concurrent map first.

spacejam:
Most of the groundwork for this has been implemented in the Snapshot refactor

sitano:
@spacejam maybe you could drop some lights onto where to look and what to do, or just point out to some other not very hard issues?

ckaran:
Can some version of this be exposed in the public API?  Here's my situation right now:

- I have multiple robots running around, each of which is running sled internally.  Each one maintains a database of what it knows other robots know (basically, every actor replicates every other actor's database).
- Communications is expensive; in some cases the robots are fully disconnected from the network, and will only be able to synchronize when they come back into communications range.  As such, latency can be on the order of hours.  Throughput is pretty bad as well, but not as bad as latency.
- Radio communications are always broadcasts; every robot can listen in on the communications of every other robot that is within range.  There is a mechanism in place for efficiently ACKing and NACKing subsequences of a complete broadcast, so if you're missing 5 KB of a 10 GB broadcast, you can ask for just that portion rather than the whole thing at once.  Think rsync, but as a group.

[...]

PR #606, opened 2019-04-01
[wip] dynamic unstable tail. closes #430

### LINKED ISSUE #430: Segment Management/Gc Refactor
There are a few assumptions and limits with how we handle Segment management.
* freed segments must wait for <#io bufs> other segments to be allocated before being given out
* we assume that we never need to reserve more memory than we have IO buffers for
* we assume that pending operations will never last longer than the time it takes to clear all IO buffers

Generally, this does not support the following needs:
* We want to be able to split huge values into chunks and write them on-log
* We want to write huge transactions over many segments

In order to do this, we need to ditch the <# io bufs> static threshold, and change the way we mark segments as Free. Right now, an EBR dropper is used to delay the transitioning of a segment to Free after writing its trailer. This is intended to allow all threads that may have held a reservation to a Segment (which has a nested EBR guard in the Reservation struct) to tell the segment accountant about any replacements before the segment is marked inactive.

This is bug-prone, overly complex, and generally a source of confusion and late-at-night-realizations that we may have a bug somewhere.

We should make the following changes:
* call deactivate segment from `IoBufs::mark_interval` when we stabilize a segment's trailer
* buffer replacements in the segment accountant until the segment is actually transitioned to Inactive

spacejam:
this branch allows us to dynamically grow or shrink the number of IO buffers in use at any time by including the stable offset in the trailer of a segment, and only deactivating segments after their LSN is passed over. we can then parallelize recovery up to the highest stabilized LSN, relying on CRDT-like LWW semantics for generating a snapshot (#587) and then perform sequential recovery in the unstable tail so that possible tears are easily detected.


spacejam:
This is going to work a bit differently to account for write batches

PR #619, opened 2019-04-06
Fix the flusher to not stop when one Tree drops

Kerollmops:
I do not know if it is the right fix to do but it appears to work now.

It seems that the Flusher is dropped multiple times (when I put log in the drop method). It is probably related to the fact that I opened a `Tree` from the `Db` and when I release it, it drops and set `shutdown` to `true`, which kills all the instances of the `Db`'s initial flusher.

Sorry if it isn't clear enough, I am currently in the train, it is a little bit hard to concentrate.

### DIFF
--- a/crates/sled/src/flusher.rs
+++ b/crates/sled/src/flusher.rs
@@ -1,14 +1,14 @@
 use std::sync::{Arc, Condvar, Mutex};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use super::*;
 
 #[derive(Debug, Clone)]
 pub(crate) struct Flusher {
-    shutdown: Arc<Mutex<bool>>,
+    waiter: Arc<Mutex<()>>,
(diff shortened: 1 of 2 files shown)


crates/sled/src/flusher.rs
```diff
@@ -94,16 +95,13 @@ fn run(
...
+        guard = sc.wait_timeout(guard, sleep_duration).unwrap().0;
     }
 }
 
 impl Drop for Flusher {
     fn drop(&mut self) {
-        let mut shutdown = self.shutdown.lock().unwrap();
-        *shutdown = true;
         self.sc.notify_all();
-        drop(shutdown);
```
Kerollmops: Do I need to `drop(self.waiter)` now? If I do not do that is this a leak?

Kerollmops:
I just tested this update with my application and spotted that the flusher hang on dropping, it seems highly related to `self.join_handle.lock().unwrap()`.

PR #620, opened 2019-04-06
Fix the flusher to not drop when a Tree is drop

Kerollmops:
This closes #619.

I just saw that you forget to put the `Flusher` into an `Arc` this way it is not dropped on the first `Tree` drop. By the way I removed the derived `Clone` on the `Flusher` to make it clear that it can't be cloned :)

By the way, would it make sense to yank the 0.21.3 version, it can make users to loose data.


Kerollmops:
I tried to add a test to ensure that we do not introduce a regression but it append to not work, even if I set `flush_every_ms` to `None` it passes.

Kerollmops:
Closed by #621

PR #621, opened 2019-04-08
fix flusher

spacejam:


spacejam:
need to make shutdown synchronous from the Drop in Context to know for sure that the flush is complete

PR #623, opened 2019-04-08
Show the error when a config cannot be loaded

Kerollmops:


Kerollmops:
Ok, the travis-ci OS killed the compilation:
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

PR #626, opened 2019-04-08
Introduce update_and_fetch/fetch_and_update methods

Kerollmops:
I do not know if it would make any sense to add this helping function to the Tree type. Should it be named like this, isn't there a specific name for this kind of function?

It could be a dangerous function because it could block the thread for an infinite time. In another hand I read somewhere that "lock-free" things are guaranteed to make progress (not like Mutex), I do not know if it can be applied to this "high level" data structure.

### DIFF
--- a/crates/sled/src/tree.rs
+++ b/crates/sled/src/tree.rs
@@ -293,6 +293,74 @@ impl Tree {
         }
     }
 
+    /// Fetch the value, apply a function to it and return the result.
+    ///
+    /// # Note
+    ///
+    /// This may call the function multiple times if the value has been
+    /// changed from other threads in the meantime.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use sled::{ConfigBuilder, IVec, Error};
+    /// let config = ConfigBuilder::new().temporary(true).build();
+    /// let t = sled::Db::start(config).unwrap();
+    ///
+    /// assert_eq!(t.update_and_fetch(b"hello", |_old| Some(vec![1])), Ok(Some(IVec::from(vec![1]))));
+    /// ```
+    pub fn update_and_fetch<K, V, F>(&self, key: K, f: F) -> Result<Option<IVec>>
+    where K: AsRef<[u8]>,
+          F: Fn(Option<&[u8]>) -> Option<V>,
+          IVec: From<V>,
+    {
+        let key = key.as_ref();
+        let mut current = self.get(key)?;
+
+        loop {
+            let tmp = current.as_ref().map(AsRef::as_ref);
+            let next = f(tmp).map(IVec::from);
+            match self.cas::<_, _, IVec>(key, tmp, next.clone())? {
+                Ok(()) => return Ok(next),
+                Err(new_current) => current = new_current,
+            }
+        }
+    }
+
+    /// Fetch the value, apply a function to it and return the previous value.
+    ///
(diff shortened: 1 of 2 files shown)


spacejam:
Yeah, this is lock-free so even though this particular thread may not ever make progress, it won't stop the whole system from making progress.

I like the idea of this function, especially because the standard library will get something similar:
https://github.com/llogiq/rust/commit/ace0a94f0881bd6b811eb160c999b67975348093

I think it should have the same name, and also return an IVec to the new value or the old value, depending on whether `fetch_and_update` or `update_and_fetch` are used.

spacejam:
In order to support these usecases more flexibly, I think we should replace some types for values so that the function will accept (or return, in this case) something that is `Into<IVec>` rather than `Vec<u8>` to reduce copies in some places

crates/sled/src/tree.rs
```diff
@@ -288,6 +288,41 @@ impl Tree {
...
+    where K: AsRef<[u8]>,
+          F: FnMut(Option<&[u8]>) -> Option<Vec<u8>>,
+    {
+        let key = key.as_ref();
+        let mut current = self.get(key)?;
+
+        loop {
+            let tmp = current.as_ref().map(AsRef::as_ref);
+            match self.cas(key, tmp, fun(tmp))? {
+                Ok(()) => return Ok(current.map(|v| v.to_vec())),
```
spacejam: this should return an IVec instead of copying the previous value to create the Vec<u8>

spacejam:
I'll cut a 0.22.1 as soon as this is merged

Kerollmops:
0.22.1 or 0.22.0 anyway no problem, Does this involve breaking changes?

crates/sled/src/tree.rs
```diff
@@ -288,6 +288,73 @@ impl Tree {
...
+    /// ```
+    /// use sled::{ConfigBuilder, Error};
+    /// let config = ConfigBuilder::new().temporary(true).build();
+    /// let t = sled::Db::start(config).unwrap();
+    ///
+    /// assert_eq!(t.fetch_and_update(b"hello", |_old| Some(vec![1])), Ok(None);
+    /// ```
+    pub fn fetch_and_update<K, V, F>(&self, key: K, mut f: F) -> Result<Option<IVec>>
+    where K: AsRef<[u8]>,
+          V: Into<IVec>,
```
spacejam: Would you mind flipping this to `IVec: From<V>` so we can make it more clear below that we're invoking that specific conversion?

crates/sled/src/tree.rs
```diff
@@ -288,6 +288,73 @@ impl Tree {
...
+          F: Fn(Option<&[u8]>) -> Option<V>,
+    {
+        let key = key.as_ref();
+        let mut current = self.get(key)?;
+
+        loop {
+            let tmp = current.as_ref().map(AsRef::as_ref);
+            let next = f(tmp).map(Into::into);
```
spacejam: same as below, please

spacejam:
This one will just be a patch bump since it's entirely new functionality

spacejam:
feel free to add a commit that bumps the sled patch version so I can just cargo publish instead of waiting for another round of tests to pass etc...

spacejam:
Good work @Kerollmops  :]

PR #627, opened 2019-04-09
accept Into<IVec> for Tree operations that provide new values

spacejam:


Kerollmops:
By the way, wouldn't you think it could be "better" to bump to 0.21.5 once #626 is merged?

spacejam:
but, now that I realized there's some breakage possibility, I'll bump to 0.22

crates/sled/src/tree.rs
```diff
@@ -210,24 +210,25 @@ impl Tree {
...
-        old: Option<&[u8]>,
-        new: Option<V>,
+        old: Option<OV>,
+        new: Option<NV>,
     ) -> Result<std::result::Result<(), Option<IVec>>>
     where
         K: AsRef<[u8]>,
-        IVec: From<V>,
+        OV: AsRef<[u8]>,
+        IVec: From<NV>,
```
Kerollmops: Why not `NV: Into<IVec>`? Just asking, no offense.

crates/sled/src/tree.rs
```diff
@@ -210,21 +210,26 @@ impl Tree {
...
         key: K,
-        old: Option<&[u8]>,
-        new: Option<Value>,
-    ) -> Result<std::result::Result<(), Option<IVec>>> {
+        old: Option<OV>,
+        new: Option<NV>,
+    ) -> Result<std::result::Result<(), Option<IVec>>>
+    where
+        K: AsRef<[u8]>,
+        OV: AsRef<[u8]>,
```
Kerollmops: Don't you remenber that it was not possible the last time I tried? 😄
I do not think that the Rust language team changed that inference behavior, sadly...

https://github.com/spacejam/sled/pull/422#discussion_r233732645

Kerollmops: Ok, we will keep these lacks of inference. I prefer that, it is more generic!

crates/sled/src/tree.rs
spacejam: because then below we can't use IVec::from(v), which is cleaner IMO than v.into() or turbofishing it

crates/sled/src/tree.rs
spacejam: cas is already sort of a rare codepath and today I'm preferring genericism

crates/sled/src/tree.rs
Kerollmops: Yeah, I agree!

PR #633, opened 2019-04-12
Replace handcrafted number conversions

Kerollmops:


spacejam:
ditto about performance here, since it's a pretty hot codepath, it would be nice to see disassemblies or benchmarks to make sure there are no performance regressions here since it's so hot

Kerollmops:
[Here is the `u64_to_arr` comparison](https://rust.godbolt.org/z/RW6VhH).

[Here is the `arr_to_u64` comparison](https://rust.godbolt.org/z/w8JgvQ).

Kerollmops:
Here are the results of a before/after stress2 run with 8 threads for 30 seconds on an Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz.

#### Before conversion rewrite (6d38f41e3086a9b3573f969dd459e9c30984b12b)

```
did 86097079 total ops in 31 seconds. 2777325 ops/s
```

[before-try-trait.txt](https://github.com/spacejam/sled/files/3077374/before-try-trait.txt)

#### After conversion rewrite (95ea50a0c464685b82a6c2a96770051c5117a4ce)

```
did 85941182 total ops in 30 seconds. 2864706 ops/s
```

[after-try-trait.txt](https://github.com/spacejam/sled/files/3077375/after-try-trait.txt)

spacejam:
this is good, but it just requires a version bump since it breaks backwards compatibility. thanks for posting the disassembly! it's always fun seeing code that looks like it should be more complex getting greatly simplified in rust by being more high level :]

Kerollmops:
Will not be OK if I make the big-endian to be a little-endian conversion? It will make the code backward compatible.

spacejam:
little-endian would be very slightly cheaper on x86 (or I assume it would be, would be interesting to see if the benchmark gets even faster) so that would also be cool!

Kerollmops:
So if I make these conversions little-endian, it could be both backward compatible and probably faster on x86? Worth it!

spacejam:
yeah, I think that's a nice way forward

Kerollmops:
Done! Do you need benchmarks now that I change big endian to little?

spacejam:
no need, thanks!

PR #635, opened 2019-04-13
Make the IVec type safer for library users

Kerollmops:
Hiding the inner enum making it impossible to read the raw inline array value therefore avoiding user mistakes (reading out of bound bytes form the `IVec::Inline` variant).

### DIFF
--- a/crates/sled/src/ivec.rs
+++ b/crates/sled/src/ivec.rs
@@ -35,25 +35,29 @@ impl<'de> Deserialize<'de> for IVec {
     }
 }
 
+const fn is_inline_candidate(length: usize) -> bool {
+    length <= CUTOFF
+}
+
 impl IVec {
-    pub(crate) fn new(v: &[u8]) -> IVec {
-        if v.len() <= CUTOFF {
-            let sz = v.len() as u8;
+    fn inline(slice: &[u8]) -> IVec {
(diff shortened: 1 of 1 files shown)


spacejam:
it makes me a bit nervous to play around so much with the various paths here without microbenchmarks around accesses / creation / serialization in such a hot codepath

Kerollmops:
Do you think running the `stress2` would be enough of a benchmark?

Kerollmops:
Here is the results of a before/after stress2 run on 30 seconds.

### Before `IVec` rewrite

```
did 85972365 total ops in 31 seconds. 2773302 ops/s
```

[before-ivec-rewrite.txt](https://github.com/spacejam/sled/files/3076450/before-ivec-rewritte.txt)

### After `IVec` rewrite

```
did 86567509 total ops in 31 seconds. 2792500 ops/s
```

[after-ivec-rewrite.txt](https://github.com/spacejam/sled/files/3076451/after-ivec-rwritte.txt)

spacejam:
in this case it's probably fine, but stress2 is quite susceptible to interference with other tasks on the machine when using the default settings. it's better to run it with --threads=1, to rm -rf the storage file first, and to disable the flusher thread. 

thanks for measuring!

Issue #533, opened 2019-02-07
clean up examples


spacejam:
the examples currently involve a lot of extra unwraps and `.to_vec` etc... and could probably be cleaned up a bit

jneem:
Which examples did you mean? I found `examples/playground` and `examples/crdt_merge_store`, but they're excluded from `Cargo.toml` and don't appear to build. Or is updating them part of the point?

spacejam:
hey jneem, mostly how there are some commented-out bits of text in the doctests that are used to convey they general point of the API but are not very clear, and don't actually compile when uncommented due to limitations in how Option and Result are implemented in the standard lib.

jneem:
Got it. I'll take a stab at improving them, and maybe also add examples for
the individual methods.

PR #636, opened 2019-04-16
Update API examples.

jneem:
This updates the API examples for `Tree`, as discussed in issue #533. The basic pattern I'm following is
```
assert_eq!(t.del(&[1]), Ok(Some(sled::IVec::from(vec![1]))));
```
That is, I'm writing out the full return type on the right hand side, including `IVec::from` for the conversion. This isn't the most concise way to write it, but I think it's nice to be as explicit as possible in the API docs. The exception is for iterator methods, where I do
```
assert_eq!(r.next().unwrap(), Ok((vec![2], IVec::from(vec![20]))));
```
That's because I'm assuming people are already familiar with the iterator API, so they already know that next returns an option.

### DIFF
--- a/crates/sled/src/ivec.rs
+++ b/crates/sled/src/ivec.rs
@@ -125,6 +125,21 @@ impl From<Vec<u8>> for IVec {
     }
 }
 
+macro_rules! from_array {
+    ($($s:expr)*) => {
+        $(
+            impl From<&[u8; $s]> for IVec {
+                fn from(v: &[u8; $s]) -> IVec {
+                    IVec::from(&v[..])
+                }
+            }
+        )*
+    }
+}
+
+from_array!(0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15);
+from_array!(16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31 32);
+
 impl Into<Arc<[u8]>> for IVec {
     fn into(self) -> Arc<[u8]> {
         match self.0 {
--- a/crates/sled/src/lib.rs
+++ b/crates/sled/src/lib.rs
@@ -9,7 +9,7 @@
 //! let t = Db::start(config).unwrap();
 //!
 //! t.set(b"yo!", b"v1".to_vec());
-//! assert_eq!(t.get(b"yo!"), Ok(Some(IVec::from(&b"v1"[..]))));
+//! assert_eq!(t.get(b"yo!"), Ok(Some(IVec::from(b"v1"))));
 //!
 //! // Atomic compare-and-swap.
 //! t.cas(
@@ -20,7 +20,7 @@
 //!
 //! // Iterates over key-value pairs, starting at the given key.
 //! let mut iter = t.scan(b"a non-present key before yo!");
-//! assert_eq!(iter.next().unwrap(), Ok((b"yo!".to_vec(), IVec::from(&b"v2"[..]))));
+//! assert_eq!(iter.next().unwrap(), Ok((b"yo!".to_vec(), IVec::from(b"v2"))));
 //! assert_eq!(iter.next(), None);
 //!
 //! t.del(b"yo!");
--- a/crates/sled/src/lib.rs
+++ b/crates/sled/src/lib.rs
@@ -3,20 +3,25 @@
 //! # Examples
 //!
 //! ```
-//! let t = sled::Db::start_default("my_db").unwrap();
+//! use sled::{ConfigBuilder, Db, IVec};
+//!
+//! let config = ConfigBuilder::new().temporary(true).build();
+//! let t = Db::start(config).unwrap();
 //!
 //! t.set(b"yo!", b"v1".to_vec());
-//! assert!(t.get(b"yo!").unwrap().unwrap() == &*b"v1".to_vec());
+//! assert_eq!(t.get(b"yo!"), Ok(Some(IVec::from(&b"v1"[..]))));
 //!
+//! // Atomic compare-and-swap.
 //! t.cas(
 //!     b"yo!",                // key
 //!     Some(b"v1"),           // old value, None for not present
 //!     Some(b"v2".to_vec()),  // new value, None for delete
 //! ).unwrap();
 //!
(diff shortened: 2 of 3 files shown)


Kerollmops:
We could probably implement `impl From<[u8; $s]> for IVec` using a macro to generate multiple array sizes.

jneem:
Good point, I've just done that. (I'm also happy to split it into a separate PR.)

crates/sled/src/lib.rs
```diff
@@ -3,20 +3,25 @@
 //! # Examples
 //!
 //! ```
-//! let t = sled::Db::start_default("my_db").unwrap();
+//! use sled::{ConfigBuilder, Db, IVec};
+//!
+//! let config = ConfigBuilder::new().temporary(true).build();
```
spacejam: even though this doctest creates garbage, I'd prefer to make the top-level example in the docs be something that won't delete itself upon Drop, since people will run it and get mad that it doesn't keep their data across process invocations even though it's kind of obvious

crates/sled/src/tree.rs
```diff
@@ -205,18 +233,17 @@ impl Tree {
...
+    /// let config = sled::ConfigBuilder::new().temporary(true).build();
     /// let t = sled::Db::start(config).unwrap();
     ///
     /// // unique creation
-    /// assert_eq!(t.cas(&[1], None as Option<&[u8]>, Some("1")), Ok(Ok(())));
+    /// assert_eq!(t.cas(&[1], None as Option<&[u8]>, Some(vec![1])), Ok(Ok(())));
     ///
     /// // conditional modification
-    /// assert_eq!(t.cas(&[1], Some("1"), Some("2")), Ok(Ok(())));
+    /// assert_eq!(t.cas(&[1], Some(&[1]), Some(vec![2])), Ok(Ok(())));
```
spacejam: I think I'd prefer to keep these as the same type to reduce confusion here

crates/sled/src/lib.rs
spacejam: (please keep the call to `Db::start_default`)

crates/sled/src/ivec.rs
```diff
@@ -125,6 +125,21 @@ impl From<Vec<u8>> for IVec {
...
+            impl From<&[u8; $s]> for IVec {
+                fn from(v: &[u8; $s]) -> IVec {
+                    IVec::from(&v[..])
+                }
+            }
+        )*
+    }
+}
+
+from_array!(0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15);
```
Kerollmops: Why splitting these macro calls? https://doc.rust-lang.org/src/core/array.rs.html#258-263
I know, I am a little picky 😄

crates/sled/src/ivec.rs
```diff
@@ -125,6 +125,21 @@ impl From<Vec<u8>> for IVec {
...
+from_array!(16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31 32);
```
Kerollmops: Why splitting these macro calls? https://doc.rust-lang.org/src/core/array.rs.html#258-263
I know, I am a little picky 😄

spacejam: I'd prefer keeping these together as well but we can fix that quickly later

spacejam:
Thanks for fixing all of these up!!! This should make learning how to use this system easier for a lot of folks :]

crates/sled/src/ivec.rs
Kerollmops: I just realized that it is probably not the right implementation needed, we would probably have prefered `impl From<[u8; $n]> for IVec {}` without the ampersand.

jneem: Can you explain why `From<[u8; $n]>` is better? I was mainly just going according to the existing `From` implementations: there's `From<&str>` and `From<&[u8]>`.

Kerollmops: I was thinking about [the `Vec` `PartialEq` impls](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-PartialEq%3C%5BB%3B%205%5D%3E) that allows to compare arrays with a `Vec` without the need to make the array be a slice (`&array[]..`).

An array of `u8` implements `Copy` anyway so it is a little bit disturbing to force the user to put an ampersand in front of its array, making the construction be `IVec::from(&[3u8; 0])` instead of the simplest `IVec::from([3u8; 0])`.

jneem: Well, `Vec` does it for both `&[u8; $n]` and `[u8; $n]`, so maybe we should also? The owned version is more convenient if your argument is `[3u8; 1]`, but the borrowed one is more convenient if your argument is `b"yo"`.

Kerollmops: You are totally right! We should probably implement `PartialEq`/`Eq` for `IVec` and `[u8; $n]`/`&[u8]` but should we do the same for `From<[u8; $n]>`? Wouldn't it be better to follow the std `Vec` implementations?

jneem: I'm not sure I follow. The std `Vec` doesn't implement `From` for any arrays, borrowed or otherwise. What do you mean by "follow the std `Vec` implementations"?

Kerollmops: I mean implementing `PartialEq`, sorry for the confusion :)

jneem: So I looked into `PartialEq`, and I don't think any more implementations are necessary, because we have `PartialEq<T>` whenever `T: AsRef<[u8]>`, which covers both borrowed and owned arrays.

Unfortunately, this route doesn't seem to be possible for `From`, because of coherence: `IVec` implements `AsRef<[u8]>`, so `impl<T: AsRef<[u8]>> From<T> for IVec` conflicts with the blanket `From<T> for T` implementation.

Kerollmops: I wonder why `Vec<T> where T: u8` do not implement `PartialEq<U> where U: AsRef<[u8]>`? 
Is it related to specialization?

PR #638, opened 2019-04-20
Improve the documentation of fetch/update tree methods

Kerollmops:


crates/sled/src/tree.rs
```diff
@@ -342,11 +368,36 @@ impl Tree {
     /// # Examples
     ///
     /// ```
-    /// use sled::{ConfigBuilder, Error};
+    /// # use sled::{ConfigBuilder, Error, IVec};
+    /// # use std::convert::TryInto;
```
spacejam: commented code still here

crates/sled/src/tree.rs
```diff
@@ -303,11 +303,37 @@ impl Tree {
     /// # Examples
     ///
     /// ```
-    /// use sled::{ConfigBuilder, IVec, Error};
+    /// # use sled::{ConfigBuilder, Error, IVec};
+    /// # use std::convert::TryInto;
```
spacejam: commented code still here

crates/sled/src/tree.rs
Kerollmops: No, it is a way to hide code from the generated documentation but it will be evaluated by the doc tests. It is a way to make the example more clear in some way.

spacejam: ohhhhh too much shell on my mind, I didn't know about this!

spacejam: one thing I've noticed from some of the download numbers is that a lot of people download sled and use it for personal stuff, but rarely publish crates that depend on it. I think part of this reason is because too many of the docs are way too long, and that some of them (like in the readme) are pseudocode that you can't just copy and paste. I think this is kind of an antipattern to have doc code that you can't just copy and paste and have work. could you remove the hiding code so people can copy and paste the example and it will work?

Kerollmops: Yeah, no problem :)

Kerollmops: done!

spacejam:
Looks good, thanks :)

Issue #524, opened 2019-02-05
refactor tree iterators


spacejam:
These are totally gross, and a source of several bugs lately

spacejam:
this is done in the `tree_avoid_merge_on_page_read` branch :]

PR #649, opened 2019-04-28
Rewrite the Iter types

Kerollmops:
I was rewriting the `Iter` type because I saw #524 and found out that the `DoubleEndedIterator` implementation was wrong, [according to the standard library](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html) forward and reverse iteration must never cross:

 > It is important to note that both back and forth work on the same range, and do not cross: iteration is over when they meet in the middle.

The current implementation seems more like some sort of cursor, if a call to `next` is followed by a call to `next_back` the iterator will return to its original position, which is not expected according to the description above. The iterator must have yielded an element from the end of the specified range when called with `next_back`.

In other words, `next` and `next_back` should consume the start or the end of the specified range and not move the same "cursor" around. Therefore having two cursors internally, one indicating the forward direction and one the backward, when one crosses the other, the iteration must stop.

Also, I was asking myself why `scan` even exists, isn't it just a `range(k..)`?
If a reverse `scan` must exists in the sense that you implement it (i.e. starts at the key but goes backward from it) it could be just something called `backward_scan(k)` which internally is just an `range(..=k)`.

### DIFF
--- a/crates/sled/src/iter.rs
+++ b/crates/sled/src/iter.rs
@@ -1,30 +1,30 @@
-use std::ops;
+use std::ops::{Bound, Bound::{Included, Excluded, Unbounded}};
 
 use pagecache::{Measure, M};
 
 use super::*;
 
-fn lower_bound_includes<'a>(lb: &ops::Bound<Vec<u8>>, item: &'a [u8]) -> bool {
-    match lb {
-        ops::Bound::Included(ref start) => start.as_slice() <= item,
-        ops::Bound::Excluded(ref start) => start.as_slice() < item,
-        ops::Bound::Unbounded => true,
+fn lower_bound_includes(bound: &Bound<Vec<u8>>, item: &[u8]) -> bool {
+    match bound {
+        Included(start) => start.as_slice() <= item,
+        Excluded(start) => start.as_slice() < item,
+        Unbounded => true,
     }
 }
 
-fn upper_bound_includes<'a>(ub: &ops::Bound<Vec<u8>>, item: &'a [u8]) -> bool {
-    match ub {
-        ops::Bound::Included(ref end) => item <= end.as_ref(),
-        ops::Bound::Excluded(ref end) => item < end.as_ref(),
-        ops::Bound::Unbounded => true,
+fn upper_bound_includes(bound: &Bound<Vec<u8>>, item: &[u8]) -> bool {
+    match bound {
+        Included(end) => item <= end.as_slice(),
+        Excluded(end) => item < end.as_slice(),
+        Unbounded => true,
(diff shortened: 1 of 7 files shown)


Kerollmops:
BTW wouldn't it be valid to implement `FusedIterator` on the Tree `Iter`?

crates/sled/src/iter.rs
```diff
@@ -1,48 +1,182 @@
...
+    let (mut page_id, mut node) = node_of_key(tree, b"", tx)?;
+
+    while let Some(next_page_id) = node.next {
+        let next_node = node_of_pageid(tree, next_page_id, tx)?;
+        page_id = next_page_id;
+        node = next_node;
+    }
+
+    Ok((page_id, node))
+}
```
Kerollmops: I would like to find a better way (e.g. binary search the biggest key) instead of doing this linked list like travel.

crates/sled/src/tree.rs
```diff
@@ -819,19 +820,56 @@ impl Tree {
...
+    /// assert_eq!(r.next().unwrap(), Ok((vec![5], IVec::from(vec![50]))));
+    /// assert_eq!(r.next().unwrap(), Ok((vec![4], IVec::from(vec![40]))));
+    /// assert_eq!(r.next().unwrap(), Ok((vec![3], IVec::from(vec![30]))));
+    /// assert_eq!(r.next(), None);
+    /// ```
+    pub fn scan<K>(&self, key: K) -> Iter
+    where
+        K: AsRef<[u8]>,
+    {
+        let key = key.as_ref().to_vec();
```
Kerollmops: Cloning the just passed byte slice reference value given by the user to make it be an `Vec` doesn't really follow the [C-CALLER-CONTROL flexibility point of the api guideline](https://rust-lang-nursery.github.io/api-guidelines/flexibility.html#c-caller-control).

Kerollmops: But I understand that you want to expose a simple and flexible API, I agree with your choice. I was just wondering if there was no other solution (i.e. `Into<Vec<u8>>`).

Kerollmops:
I do not think that the CI test not passing is related to the work I have done. It is related to the `Op::GetLt` operation that do not output the same result than the reference `BtreeMap`.

Additionally, when I run this test on my computer it works well.

spacejam:
Your work on this is really helpful because it made me realize there are even more simplifications that we can do here! I'm currently doing a huge refactor of the tree related code though, as part of transaction prep, so this might not be easy to merge. I'm going to take some of the ideas here though while simplifying the iterator

PR #650, opened 2019-05-01
Add OpenDb trait and impl it for AsRef<Path> #521

spebern:


spacejam:
Thanks for adding this! Could you change `start_default` to `open_db` to pass more specific information to readers of code, and also reexport it from `src/lib.rs`?

Kerollmops:
Does this trait addition really add value to the library?

It doesn’t really reduce the boilerplate to open a database (`Db::open_default(path)?` vs `path.open_db()?`) and neither reduce the number of imports (`sled::Db` vs `sled::DbOpen`). But I think it reduces the expressiveness of the expression as it hides the output type.

spacejam:
These are fair points. To be honest, I was mostly just interested in this
because it felt cute to me when I imagine seeing it in a README, but it
does add a layer of indirection that may outweigh any feeling of lightness.
I'll ask a few more folks about how it makes them feel.

Lurkers out there: anyone have an opinion about this? This could be one of
the asthetic front doors to using sled, and I want to make sure it is
inviting.

Maybe it's good to have as few front doors as possible, too...

ehiggs:
This is cute code but reduces the greppability. Instead of being able to immediately see where open_db comes from, we have to infer that it's part of OpenDb. I think it's not ideal in the same way that using `use sled::*;` wouldn't be ideal in a code base unless there's a really good reason.

Note this is distinct from traits like Read where indeed you need to import `Read` in order to use the trait on a struct, but those traits pull their weight.

spacejam:
Thanks for the feedback! Yeah, I think in a case like this the right decision is to resist cuteness and maximize greppability. I'll close this and the backing issue.

@spebern thanks for implementing this and triggering the valuable feedback to come in :D Unfortunately the code will be sacrificed, but your effort allows the underlying issue to be closed. Thank you for spending the time to help the project, and I hope we're able to collaborate more going forwards!

PR #659, opened 2019-05-12
Make Tree opening more configurable

Kerollmops:
This pull request will introduce better ways to open Trees allowing users to specify:
  - if a tree that doesn't exist must be created or not
  - if a tree must not exist to be created

Introducing a new `TreeOpenOptions` type will permit more future features like:
  - a custom merge operator by Tree
  - a truncate open option to clear a tree if it does already exist
  - compression settings by tree


spacejam:
I think this is good! Over time it would be nice to add other fs::OpenOptions-like things here

spacejam:
Closing this because per-tree merge operators are now supported on master

Kerollmops:
Wow 🎉 That's so cool! Thank you for your work Tyler!

Issue #661, opened 2019-05-19
unclear how keys are orderd for multi byte keys 


dskleingeld:
Version: 0.24.1

I can not find in what order keys are returned by `get_gt()`, `get_lt()` and range when I am using keys of more then one byte. 

For example if I created keys using `i32::to_le_bytes()`, will the keys be returned (by range) in the same order as their signed integer representation? Clearly the order can not follow the integer representation using `i32::to_le_bytes()` and `i32::to_be_bytes()`. 

Should I not create keys this way? I am trying to store a number of events (values) by timestamp (key). 

I would be happy to try and improve documentation once I understand how the order is determined.

spacejam:
it's simple lexicographic order, as matches the Ord implementation for the canonical type:
```
[] < [0] < [255] < [255, 0] < [255, 255] ...
```

If you're trying to retain the lexicographic ordering of numerical types, use big endian.

In general, scanning forward (`[] -> [x, ...]`) is faster than backwards, and get_gt uses the forward iterator under the hood. So, if you need to do a lot of scanning, keep this in mind.

PR #665, opened 2019-05-23
add order explanation and example to `Tree::get_gt` for multi byte keys

dskleingeld:
Adds the explanation from @spacejam from #661 to the `get_gt` function. I don not know if this is useful for you, if it is not feel free to close this PR. If you would like I will expand this PR to also extend the documentation for `get_lt` and `range` in a similar way.


spacejam:
Thanks for cleaning up the docs, @dskleingeld  :]

PR #670, opened 2019-05-29
Avoid pointer chasing in the PageTable

spacejam:


crates/pagecache/src/ds/pagetable.rs
```diff
@@ -40,24 +28,32 @@ fn split_fanout(i: u64) -> (u64, u64) {
...
 }
 
-impl<T: Send + 'static> Default for Node1<T> {
-    fn default() -> Node1<T> {
-        let children = rep_no_copy!(Atomic::null(); FANOUT as usize);
-        Node1 { children }
+impl<T: Send + 'static> Node1<T> {
+    fn new() -> Box<Node1<T>> {
+        let mut node: Box<Node1<T>> =
+            unsafe { Box::new(std::mem::uninitialized()) };
```
spacejam: we need to allocate the Node_ directly on the heap since it's 2mb and otherwise it could cause a stack overflow

Kerollmops: That's exactly the type of unsafe block that [the `box` syntax](https://doc.rust-lang.org/unstable-book/language-features/box-syntax.html) will remove :)

Kerollmops:
I found that the fix you talk above to avoid stack overflows doesn't actually work. As the [Box::new constructors explains](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.new), an allocation is done on the heap **and then** the value is copied into it. So the unsafe call to `mem::unitialized` has a stack size of `FANOUT * Atomic<usize>`.

The only way I see to avoid a stack overflow is to change the `Node1` type to hold an `Box<[Atomic<Node2<T>>; FANOUT]>`. The [`Box<[T]> implement Default`](https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-Default) and [`Atomic implement Default`](https://docs.rs/crossbeam-epoch/0.7.1/crossbeam_epoch/struct.Atomic.html#impl-Default) too, therefore it will be possible initialize it without passing to stack, I hope.

PR #672, opened 2019-06-08
Fix the stack overflow in the PageTable creation

Kerollmops:
This pull request override the previous attempt made in https://github.com/spacejam/sled/pull/670.

I made this PR in two commits, the first shows a fix that use entirely safe Rust code but do not keep the types original type `Box<[_; FANOUT]>` and instead use boxed slices `Box<[_]>`.

The second commit use the unsafe [std alloc](https://doc.rust-lang.org/std/alloc) functions to allocate an array and I initialize it with `Atomic::null`s. This second commit as the advantage to keep the strong initial types `Box<[_, FANOUT]>`.

### DIFF
--- a/crates/pagecache/src/ds/pagetable.rs
+++ b/crates/pagecache/src/ds/pagetable.rs
@@ -28,22 +28,59 @@ fn split_fanout(i: u64) -> (u64, u64) {
 }
 
 struct Node1<T: Send + 'static> {
-    children: Box<[Atomic<Node2<T>> /* FANOUT */]>,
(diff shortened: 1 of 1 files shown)


spacejam:
if we use Boxes we may as well just use a Vec and avoid all unsafe, but this still does a double pointer chase. We can skip a lot of the casting and pointer math here by avoiding the inner Box I think. I'll give it a swing now

spacejam:
what do you think about https://github.com/spacejam/sled/pull/673 ?

Kerollmops:
Yeah, that’s better! ^^

PR #673, opened 2019-06-08
Directly allocate nodes on the heap

spacejam:


crates/pagecache/src/ds/pagetable.rs
```diff
@@ -37,8 +44,14 @@ struct Node2<T: Send + 'static> {
...
-            unsafe { Box::new(std::mem::uninitialized()) };
+        let size = size_of::<Node1<T>>();
+        let align = align_of::<Node1<T>>();
+        let mut node = unsafe {
+            let layout = Layout::from_size_align_unchecked(size, align);
+            let ptr = alloc(layout) as *mut Node1<T>;
+            Box::from_raw(ptr)
+        };
+
         for i in 0..FANOUT as usize {
```
Kerollmops: I would have preferred directly iterating over `node.children` instead of using `FANOUT`, it would have been better defensive programming oriented :)

Same for the similar loop below.

spacejam: Good idea!

spacejam: realized we can actually just use alloc_zeroed and skip that part completely https://github.com/spacejam/sled/pull/674

Kerollmops: I do not think that it is safe to do that, I hesitated to zero the allocation and do not initialize it with `Atomic::null` but I think that LLVM will remove the initialization loop as it sees that it is just a bzero.

The safest way could be to zero the allocation **and** to initialize it, this way LLVM will remove the useless initialization loop.

BTW [the `Atomic::null` constructor will be const](https://docs.rs/crossbeam-epoch/0.7.1/src/crossbeam_epoch/atomic.rs.html#153-176) when stabilized and that will make this part even easier to be pruned by LLVM

Kerollmops: I would have loved a @RalfJung advice here :)

RalfJung: I'm happy to help, but please formulate a self-contained question. :)

Kerollmops: We would like to allocate directly on the heap, this allocation will be an array of [`crossbeam-epoch::Atomic`](https://docs.rs/crossbeam-epoch/0.7.1/crossbeam_epoch/struct.Atomic.html) types. We would like all these `Atomic`s to be initialized to [`Atomic::null`](https://docs.rs/crossbeam-epoch/0.7.1/crossbeam_epoch/struct.Atomic.html#method.null), is it safe to call [`alloc_zeroed`](https://doc.rust-lang.org/stable/std/alloc/fn.alloc_zeroed.html) and use it directly or is it mandatory to loop on the just zeroed allocated array and set each element to `Atomic::null`?

spacejam: FWIW ASAN seems happy. But I'm curious if my assumption that alloc_zero "counts" as initialization for an array of null pointers is wrong. Does that sound dubious to you, @RalfJung?

spacejam: And, since it's unclear, [here](https://github.com/spacejam/sled/pull/674) is the actual code we're talking about

RalfJung: Yes, `alloc_zero` "counts as" initializing all integers and floating points to 0 and raw pointers to NULL. You are relying here on the fact that there are no other types involved in this! So if `Atomic` ever changes  the fields you might get into trouble.

Doesn't seem likely that they will, though...

RalfJung: I am curious why you might think that this is not okay? I can see people being paranoid over various things but zero initialization was not on my list for that. (But to be clear, I am always happy when people are cautious around this stuff instead of playing it fast and loose!)

Kerollmops: To be clear I was dubious about initializing atomic types this way and also was wondering if it was safe if there is changes in the `Atomic` struct in the future.

Now that you validate that it is safe to do so, we can now merge this pull request.

Thank you for your reply :)

spacejam: Thanks, @RalfJung :] Always better to be thorough since mistakes in databases can leave people permanently impacted.

@Kerollmops it's possible that crossbeam-epoch could add additional fields to `Atomic` going forward. It currently overloads an `AtomicUsize` with both the pointer and "tag bits" which may store a few bits of information in the low bits that are impossible to occupy due to alignment. I already checked that this is 0 if the tag bits are unused. But if there are any new fields which are not 0-friendly, this will become fucked. 

@stjepang have you thought about adding anything additional to the `Atomic` type?

jeehoonkang: I'm not sure Crossbeam can commit to a representation of `Atomic<T>` at this stage.  How about checking if the representation of `Atomic::null()` is actually zero, dynamically (`debug_assert!`) or statically (`const_assert!`)?

RalfJung: `Atomic::new` also uses non-atomic writes to initialize to 0, so the atomicity is no concern here.

> How about checking if the representation of Atomic::null() is actually zero, statically (const_assert!)?

Depending on how the representation changes, the check might be UB (namely, comparing padding bytes is UB)... so if possible I'd advise for doing this statically.

Kerollmops: The `Atomic::null` method is [currently not const](https://docs.rs/crossbeam-epoch/0.7.1/src/crossbeam_epoch/atomic.rs.html#170-176). I think it is probably overkill, though.

Another solution could be to initialize the array elements to `Atomic::null` call in a for loop. This way a good compiler will optimize away the loop if it see redondancy (two bzero one after the other).

PR #680, opened 2019-06-12
Remove SegmentTrailer. Closes #645.

### LINKED ISSUE #645: remove segment trailer
this has not been necessary since padding was introduced, and it can double the write latency for a segment because it has its own fsync. 

* we can move max_stable_lsn into the header. 
* we can keep LSN on padding messages which keeps us resilient against tears at the previous pad location

spacejam:


crates/pagecache/src/iobuf.rs
```diff
@@ -128,15 +128,19 @@ impl IoBufs {
             next_lid
         );
 
+        // we want stable to begin at -1 if the 0th byte
+        // of our file has not yet been written.
+        let stable = next_lsn - 1;
```
Kerollmops: You should use the [`wrapping_sub`](https://doc.rust-lang.org/std/primitive.isize.html#method.wrapping_sub) function to avoid Rust to panic on overflowing arithmetic operations.

Kerollmops: If you always want to use the `-1` value as a marker, obviously.

spacejam: it's an isize and we want next_lsn of 0 to be -1

Kerollmops: Ahha! Wooops...

PR #692, opened 2019-06-16
fix up event log and turn it back on during tests

spacejam:
Remove segment-related event log items, as it's not deterministic enough to be useful. Move page-related event logging to the pagecache itself. Update travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,8 +64,7 @@ script:
                osx)
                  pushd tests &&
                  cargo check &&
-                 cargo check --features=all &&
-                 cargo test tree --release
+                 cargo test tree
                  ;;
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -64,8 +64,7 @@ script:
                osx)
                  pushd tests &&
                  cargo check &&
-                 cargo check --features=all &&
-                 cargo test tree --release
+                 cargo test tree
```
Kerollmops: Did you remove `--all-features` deliberately or did you forget to add it?

spacejam: deliberate, and I got rid of all of the features on the tests subcrate

PR #702, opened 2019-06-29
Introduce the Tree scan_prefix method

Kerollmops:
I would like to know what your thoughts about the fact that my method returns and unnamed `Iterator` type. Do you think it could be better to create a concrete Iterator type?

Note that [the only traits that a `TakeWhile` transfert](https://doc.rust-lang.org/std/iter/struct.TakeWhile.html) is the `Iterator` and the `FusedIterator` ones, it is not possible to have a `DoubleEndedIterator` using `TakeWhile` therefor we do only loose the `Debug`, `FusedIterator` and `Clone` traits impls by exposing an unnamed type this way.

### DIFF
--- a/crates/sled/src/tree.rs
+++ b/crates/sled/src/tree.rs
@@ -817,6 +817,47 @@ impl Tree {
         }
     }
 
+    /// Create an iterator over tuples of keys and values,
+    /// where the all the keys starts with the given prefix.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use sled::{ConfigBuilder, Db, IVec};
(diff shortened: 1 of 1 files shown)


spacejam:
Maybe something like this?
```
let mut upper = prefix.clone();
while !upper.is_empty() {
    let last = upper.pop().unwrap();
    if last < u8::max_value() {
        upper.push(last + 1);
        return self.range(prefix..&upper);
    }
}
self.range(..)
```

Kerollmops:
Yeah, that's much cleaner I agree! :)

spacejam:
Looks good, nice feature :]

PR #703, opened 2019-06-30
Use next_back in the Iterator last method

Kerollmops:


spacejam:
Good catch! Thanks :]

PR #706, opened 2019-07-15
cargo/pagecache: update to fail 0.3

lucab:
Code generation by the underlying library is now disabled by default and
can be opted-in (and re-exported) via the `fail/failpoints` feature.

Ref: https://github.com/pingcap/fail-rs/blob/v0.3.0/CHANGELOG.md


spacejam:
Awesome, thanks @lucab  :D

PR #708, opened 2019-07-15
Use fail 0.3 in failpoint property tests

spacejam:


lucab:
My bad, I missed the test-runner.
You can directly use [`FailScenario`](https://docs.rs/fail/0.3.0/fail/struct.FailScenario.html), it internally takes care of cleaning up leftovers in case of previous panics.

PR #725, opened 2019-07-24
optimize startup

spacejam:
cc @mxinden


spacejam:
This cuts initial startup + shutdown time down quite a lot from the last optimization, most of the time just doing one fsync on Db::Drop, which should not be avoided. Most of the remaining latency that could be cut off is coming from the fact that we use large pagetable nodes that take a while to Drop, but there may be more ways to get through this faster

PR #726, opened 2019-07-24
don't scan through empty pagetable entries on Drop because it's assumed to be dense

spacejam:


spacejam:
@mxinden and this should get rid of almost all of the pagetable destruction latency

spacejam:
There are some other areas that could still be optimized but it's sooo much better than before

[flamegraph.svg.zip](https://github.com/spacejam/sled/files/3428911/flamegraph.svg.zip)

Kerollmops:
Do you have a before after svg? :)

spacejam:
@Kerollmops unfortunately I deleted them already, but the difference is really clear when just starting and dropping a non-temporary database with a unique name 5000 times

PR #730, opened 2019-07-26
Deprecate Tree::del and replace it by Tree::remove

### LINKED ISSUE #729: rename del to remove, add deprecated pass-through from del to remove
to make the Tree api more similar to the std map collections, we should rename del to remove and add a deprecated pass-through method so no code breaks for now. This should all happen in tree.rs

Kerollmops:
Closes #729.

### DIFF
--- a/crates/sled/src/tree.rs
+++ b/crates/sled/src/tree.rs
@@ -42,7 +42,7 @@ impl<'a> IntoIterator for &'a Tree {
 /// assert_eq!(iter.next().unwrap(), Ok((IVec::from(b"yo!"), IVec::from(b"v2"))));
 /// assert_eq!(iter.next(), None);
 ///
-/// t.del(b"yo!");
+/// t.remove(b"yo!");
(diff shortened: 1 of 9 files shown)


crates/sled/src/tree.rs
```diff
@@ -198,12 +198,28 @@ impl Tree {
     /// assert_eq!(t.del(&[1]), Ok(Some(sled::IVec::from(vec![1]))));
     /// assert_eq!(t.del(&[1]), Ok(None));
     /// ```
+    #[deprecated(since = "0.24.2", note = "replaced by `Tree::remove`")]
```
Kerollmops: I chose `0.24.2` but it is up to you to tell me the version where the deprecation begins.

spacejam: That's good! The next version to be published will either be 0.25 or 0.666 depending on if I want to do one before beta.

spacejam:
Looks good, thank you!

PR #731, opened 2019-07-26
Deprecate Tree::set and replace it by Tree::insert

### LINKED ISSUE #728: rename set to insert, add deprecated pass-through from set to insert
In tree.rs, we should rename the Tree::set method to Tree::insert to be more in-line with the std map operations. We should also add a deprecated set method so that no code breaks until we remove it in a future release.

Kerollmops:
Closes #728.

### DIFF
--- a/crates/sled/src/tree.rs
+++ b/crates/sled/src/tree.rs
@@ -26,7 +26,7 @@ impl<'a> IntoIterator for &'a Tree {
 /// use sled::{Db, IVec};
 ///
 /// let t = Db::start_default("db").unwrap();
-/// t.set(b"yo!", b"v1".to_vec());
+/// t.insert(b"yo!", b"v1".to_vec());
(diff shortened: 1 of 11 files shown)


crates/sled/src/tree.rs
```diff
@@ -72,16 +72,38 @@ impl Tree {
     /// assert_eq!(t.set(&[1,2,3], vec![0]), Ok(None));
     /// assert_eq!(t.set(&[1,2,3], vec![1]), Ok(Some(IVec::from(&[0]))));
     /// ```
+    #[deprecated(since = "0.24.2", note = "replaced by `Tree::insert`")]
```
Kerollmops: I chose `0.24.2` but it is up to you to tell me the version where the deprecation begins.

spacejam:
Looks good! I just resolved a merge conflict and I'll merge as soon as it's green

PR #742, opened 2019-08-05
replace db::start_default with db::open

sitano:
I don't know how everybody is gonna like this. This replaces `db::start_default` with `db::open` to be easier and something that everyone expects from the API, similar to a `std::fs::File`. 

Now, I don't like that there is `db::start(config)` left and it's inconsistent with the `db::open(path)`. I would replace it with something like `db::init(config)` or `db::open_config(config)`. What do you think?

It seems like there is still no func overloading in Rust.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ An (alpha) modern embedded database. Doesn't your data deserve an (alpha) beauti
 ```rust
 use sled::Db;
 
-let tree = Db::start_default(path)?;
+let tree = Db::open(path)?;
 
 // set and get
 tree.insert(k, v1);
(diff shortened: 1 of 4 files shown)


spacejam:
Thanks Ivan! Don't worry about these particular tests failing, as there are a few hiccups as I move over to azure and am testing things on windows for basically the first time in a serious way. Really the only important tests to pass are the sanitizers, doctest, and the linux-stable tests until I can really debug that issue on windows (I think it's because a mutex is being taken out before doing any IO on a file, but maybe that's not true). 

I also have been thinking about the config issue, but think it might be better to move it more toward `std::fs::OpenOptions` where after setting options you just call `open` on it. But there are a few issues that need to be worked through before we make that change, since right now we heavily rely on the ability to open the same DB multiple times over the course of a single test, and that needs to be addressed before `start` can be cleaned up as nicely.

PR #745, opened 2019-08-06
add dyn for boxed traits

sitano:
latest rustc is unhappy about missing dyn`s before boxed traits, and
there are some places where the compiler can auto infer lifetime context.


spacejam:
nice cleaups, thanks Ivan!

PR #757, opened 2019-08-14
Avoid encoding keys that are already encoded

spacejam:


spacejam:
this starts to matter if keys tend not to share a very long prefix, are over the inline limit of ivec, and tend to be rewritten

PR #761, opened 2019-08-15
Fix some of the clippy advices

### LINKED ISSUE #760: pedantic clippy cleanup
There are many issues when running `cargo clippy -- -W clippy::pedantic` from the root. Most of them are quite easy to fix (please ignore cyclometric complexity issues).

Please help us clean some of them up! Even a single one is helpful!

Kerollmops:
A little bit of the work needed for #760.


spacejam:
wow, so much cleanup! great work, @Kerollmops :D

PR #762, opened 2019-08-17
Some fixes for Clippy pedantic lints

divergentdave:
Here's a raft of fixes for #760. I mostly skipped lints about sign loss/truncation on casting, function arguments that could be passed by reference instead, and binding names.


spacejam:
Looks great, thanks for this huge cleanup @divergentdave :D

PR #763, opened 2019-08-18
simple multi-Tree serializable transactions

spacejam:
this is an implementation of a simplified transaction API that can be used for single Trees or tuples of Trees for multi-Tree transactions.

in-scope:
* nice API
* serializable

future work:
* scans
* optimistic concurrency control
* MVCC

todo:

- [x] properly materialize conflicts by comparing current values to read values during validate phase after staging
- [x] migrate `Batch` to use this underneath (thus adding support for multi-Tree batches)
- [ ] use fine-grained locking

### DIFF
--- a/crates/pagecache/src/pagecache.rs
+++ b/crates/pagecache/src/pagecache.rs
@@ -96,6 +96,9 @@ where
 /// before any of the atomic batch can be partially recovered.
 ///
 /// Must call `seal_batch` to complete the atomic batch operation.
+///
+/// If this is dropped without calling `seal_batch`, the complete
+/// recovery effect will not occur.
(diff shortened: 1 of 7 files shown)


Kerollmops:
Ahah! Funny because I just tried 1h ago to expose to the user the same kind of API on the `sled::Tree` using the code I found in the [`Batch::apply`](https://docs.rs/sled/0.25.0/src/sled/batch.rs.html#32-48) method :)

spacejam:
Skipping fine-grained locking for now so that people can just use the API and take advantage of the feature :]

Feedback welcome!

theduke:
I've been waiting for transactions to materialize, so this is awesome. 
Thanks for your work @spacejam .

PR #765, opened 2019-08-19
Clippy: Pass params as references where possible

Jibbow:
Some more work for #760. 

Remarks:

It is also possible to change 
```
pub fn cas<K, OV, NV>(
    &self,
    key: K,
    old: Option<OV>,
    new: Option<NV>
)
```
to have the `old` param as a reference. However, it's a public function so I didn't touch it.

### DIFF
--- a/crates/pagecache/src/util.rs
+++ b/crates/pagecache/src/util.rs
@@ -23,7 +23,7 @@ pub(crate) fn u32_to_arr(number: u32) -> [u8; 4] {
     number.to_le_bytes()
 }
 
-pub(crate) fn maybe_decompress(buf: &[u8]) -> std::io::Result<Vec<u8>> {
+pub(crate) fn maybe_decompress(buf: Vec<u8>) -> std::io::Result<Vec<u8>> {
     #[cfg(feature = "compression")]
     {
         use std::sync::atomic::AtomicUsize;
@@ -57,5 +57,5 @@ pub(crate) fn maybe_decompress(buf: &[u8]) -> std::io::Result<Vec<u8>> {
     }
 
     #[cfg(not(feature = "compression"))]
-    Ok(buf.to_vec())
+    Ok(buf)
 }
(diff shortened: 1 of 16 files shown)


crates/pagecache/src/util.rs
```diff
@@ -57,5 +57,5 @@ pub(crate) fn maybe_decompress(buf: Vec<u8>) -> std::io::Result<Vec<u8>> {
     }
 
     #[cfg(not(feature = "compression"))]
-    Ok(buf)
+    Ok(buf.to_vec())
```
spacejam: This forces all data going through this path to be copied when compression is disabled. Please keep this argument as requiring an owned `Vec` so this copy is avoided when compression is disabled.

Jibbow: Oh yes, sure! I wasn't aware of this as I am relatively new to Rust. Thanks for the hint :+1:

Jibbow:
It's really just a small contribution, but its my first one _ever_ to a public project :sweat_smile: 
I thought this was a good issue to get started with this awesome project.

I'd appreciate any feedback on the PR :blush:

crates/pagecache/src/util.rs
spacejam: In your usage of it, it's actually fine because it only gets called if compression is enabled anyway (and clippy was right to point it out as a possible optimization), but this makes it a little more defensive in case something changes over time.

The main thing to be careful of in a database is anything that takes a buffer and makes a duplicate of it. Avoiding this is the single biggest performance trick in database engineering lol

spacejam:
Thanks for all of this clean-up, @Jibbow :D Looks great!

spacejam:
For cas, I'll juggle some constraints around now to try to make it extra flexible! It might be a little tricky because of all of the type stuff going on, but I'll tag you on the PR in case it's interesting to you!

Jibbow:
Oh yes, that would be great! :grin: Thank you!

spacejam:
@Jibbow after juggling some type constaints I realized it might be a bit trickier than I imagined to do this (Just making the old value AsRef<Option<OV>>, OV: AsRef...) I went down a rabbit hole that ended up not making things very nice to work with. I think the current situation isn't too bad, since it's just an Option of a reference, and maybe clippy pedantic isn't to be listened to in this case.

PR #773, opened 2019-08-22
Fix All CRLF line separators into LF.

flame4:
Hi， I found there's some file make crlf instead of lf as line separators so i fix them.
they are all not rust file.


spacejam:
Thanks for the cleanup, @flame4! I just merged the current master into this and I'll merge your pull request as soon as tests pass! (they are not relevant for the specific changes you pushed but it's the github policy I set for the project overall)

PR #776, opened 2019-08-23
remove 'a, replace deprecated calls, print last error if failed to lock db on open

sitano:


crates/pagecache/src/config.rs
```diff
@@ -337,8 +337,8 @@ impl ConfigBuilder {
                         return Err(Error::Io(std::io::Error::new(
                                     std::io::ErrorKind::Other,
                                     format!(
-                                    "could not acquire appropriate file lock on {:?}",
-                                    path
+                                    "could not acquire appropriate file lock on {:?}: {:?}",
+                                    path, lock_res.unwrap_err()
```
Kerollmops: Would be better is you replace the `if lock_res.is_err()` by an `if let Err(e) = lock_res` and remove the `unwrap_err()`. :)

sitano: Totally agree!

sitano: Done!

sitano:
I suppose `io::ErrorKind::Other` may be replaced with `io::ErrorKind::WouldBlock` to allow caller to understand that the error is not permanent and may be retried. WouldBlock may partially reflect the nature of a failure for the try_lock.

sitano:
I don't understand how it can be that for a new temporary database just newly created file fails exclusive lock acquisition:

```
IO error: could not acquire appropriate file lock on "C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\pagecache.tmp.9115935938206236672\\db": Os { code: 33, kind: Other, message: "The process cannot access the file because another process has locked a portion of the file." }', crates\pagecache\src\config.rs:202:13```

from sled code it follows that there is no other process that must be accessing it right after creation. May be its windows specific behavior of the API, or some other process...

spacejam:
Thanks! Will merge after tests pass again on new master. 

Yeah, the windows lock acquisition issue is a bit confusing for me as well, and am expecting I might need to dust off my old windows machine to get more insight.

sitano:
https://gitlab.gnome.org/GNOME/glib/merge_requests/746

it is possible that this ERROR_LOCK_VIOLATION is just a temporary situation and worth a retry.

spacejam:
I'm not certain about ErrorKind::WouldBlock because it might be the case that another database is going to hold a conflicting lock more-or-less permanently on a system. But I do think it might be nice to be clear about which lock was unable to be attained.

Issue #685, opened 2019-06-14
on linux, detect if a cgroup memory limit is in place and bound cache based on that


spacejam:
issue a warning log message as well

sitano:
1. read /sys/fs/cgroup/memory/memory.limit_in_bytes as in https://github.com/scylladb/seastar/blob/7dce74bff758157dc4b3072483a089c178f59504/src/core/resource.cc#L110
2. fallback to https://github.com/dotnet/coreclr/blob/master/src/pal/src/misc/cgroup.cpp#L385-L428

https://github.com/dotnet/corefx/issues/35638
https://lwn.net/Articles/432224/

sitano:
@spacejam do you think it must be:
- 100% from cgroup mem limit
- 80% from cgroup mem limit
- max(1gb (current default), cgroup mem limit) 
?

spacejam:
closed by #789

PR #778, opened 2019-08-26
limit cache size by cgroup memory limit

sitano:
Implementation of https://github.com/spacejam/sled/issues/685: a cgroup memory limit is in place and bound cache based on that.

Do we need tests here? (at least if cgroup sys fs is present on the host)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,29 +71,23 @@ steps:
     displayName: doctests
     condition: eq( variables['testKind'], 'doctests' )
   - script: |
-      echo "cross build" &&
-      echo "https://github.com/rust-lang/cargo/issues/4753" &&
-      rustup target add aarch64-fuchsia &&
-      rustup target add aarch64-linux-android &&
-      rustup target add i686-linux-android &&
-      rustup target add i686-unknown-linux-gnu &&
-      rustup target add x86_64-pc-windows-gnu &&
-      rustup target add x86_64-linux-android &&
-      rustup target add x86_64-fuchsia &&
-      pushd crates/sled      && cargo check --target aarch64-fuchsia && popd &&
-      pushd crates/sled      && cargo check --target aarch64-linux-android && popd &&
-      pushd crates/sled      && cargo check --target i686-linux-android && popd &&
-      pushd crates/sled      && cargo check --target i686-unknown-linux-gnu && popd &&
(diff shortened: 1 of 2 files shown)


sitano:
It's ok we print WARN to stdout?

spacejam:
@sitano I would use the `warn!` logging macro there instead of `println!`. Rust also has an `eprintln!` macro for printing error messages to stderr but I think in this case it's fine to just use the built-in logger.

crates/pagecache/src/config.rs
```diff
@@ -688,3 +697,20 @@ impl Config {
...
 }
+
+/// See the Kernel's documentation for more information about this subsystem, found at:
+///  [Documentation/cgroup-v1/memory.txt](https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt)
+#[cfg(target_os="linux")]
+fn get_cgroup_memory_limit() -> u64 {
+    std::fs::File::open("/sys/fs/cgroup/memory/memory.limit_in_bytes")
+        .and_then(read_u64_from)
+        .unwrap_or(0)
+}
```
spacejam: In my mind I imagined a more annoying implementation than this, but this looks nice! The one change I would make to this is to return an Option (by replacing `.unwrap_or(0)` with `.ok()` to convert the result to an Option, and then in the caller above instead of checking for if it's above 0, using an `if let Some(mem) = get_cgroup_memory_limit() { ...`

sitano: > I imagined a more annoying implementation 

What do you think of? I can abstract it more, reuse `cgroup-rs` or fallback to checking rlimit/as?

spacejam: I think the cgroup stuff itself is good!

azure-pipelines.yml
```diff
@@ -71,29 +71,23 @@ steps:
...
+      for crate in $crates; do
+      for target in $targets; do
+        pushd crates/$crate
+        echo "setting up $target..."
+        rustup target add $target
+        echo "checking $target..."
+        cargo check --target $target
+        popd
+      done
+      done
```
spacejam: could you put this in a separate PR? It causes a conflict with a PR I currently have open which migrates more to using github actions, and I'd like to keep things separate in PR's to make small changes easier to merge.

sitano: yes, easily

PR #788, opened 2019-08-29
fix cross-compile script to print a platform which broke the build

sitano:
and reuse bash for loops to optimize code boilerplate


spacejam:
Looks good! I'll merge this when the tests go green after merging master in!

PR #789, opened 2019-08-29
limit cache size to the cgroup memory limit on linux

sitano:
Implementation of https://github.com/spacejam/sled/issues/685: limit cache size to the cgroup memory limit on linux. It does not contain tests tho.


spacejam:
Looks good! Thanks for this work :) Testing for this might involve a bit of complexity that isn't clearly worthwhile right now, but as long as it has worked once in a cgroup that's fine for the time being

PR #791, opened 2019-08-30
Add check to transaction test after writers finish

divergentdave:
This adds one more assert to the `concurrent_tree_transactions` test, after all the reader and writer threads have joined. It checks that the entries have returned to their original values (since they get swapped an even number of times). If any of the swapping transactions gets lost, and their durability isn't maintained, this assert will catch the issue.


spacejam:
This is a cool test, simple and important :) I'll merge master in after your PR renaming Event variants, and this is up next to merge!

PR #796, opened 2019-09-03
fix some ds mod warns and refactor it a little #795

sitano:
https://github.com/spacejam/sled/issues/760

- fixed some clippy for `ds` mod
- fixed cycle dependency for `ds` mod
- make some ds mod structs names easier
- simplify some code
- removed Node{1,2} drop code dup
- remove PageId abstraction leak to the DLL, LRU
- used inlined `safe_usize` helper instead of `usize::try_from.unwrap`
- add a script that helps to check cross-compilation locally
- split_fanout made more natural to the target types and moved closer to the traversal

### DIFF
--- a/crates/pagecache/src/config.rs
+++ b/crates/pagecache/src/config.rs
@@ -204,14 +204,14 @@ impl ConfigBuilder {
             now + seed
         };
 
-        return if cfg!(target_os = "linux") {
+        if cfg!(target_os = "linux") {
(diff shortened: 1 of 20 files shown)


sitano:
@spacejam reworked issues with usize::try_from. you are welcome to review it further)

spacejam:
Looks great! Thanks for all of this nice work, @sitano :)

PR #798, opened 2019-09-04
[wip] [test] catch full lock time conflict

sitano:


sitano:
a version of code at b0da128 failed at https://dev.azure.com/tylerneely/sled/_build/results?buildId=220 est src\tree.rs - tree::Tree::remove (line 346) ... FAILED with file exists "C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\pagecache.tmp.9353121864995569664\\db" before the test', crates\pagecache\src\config.rs:342:13 which means there was a database random name collision while executing tests of tree mod

PR #802, opened 2019-09-06
[WIP] typed Trees

spacejam:


Kerollmops:
I hope the typed `sled::Tree` is not mandatory or there is a way to let users put raw bytes *without* calling `bincode`, in my usage of `sled::Tree` I made my types follow of a specific memory representation to be sure I can use fast `zerocopy` de/serialization.

spacejam:
@Kerollmops it's quite important to allow the default Tree<IVec> to continue having the same writepath with no extra copies or passes through serialization

spacejam:
I'll probably change a few things around as far as memory layout goes, like making a generic Inlinable type that is like an IVec but for anything that has an alignment that divides 32, up to a size of 28 or something. In general, IVec has an alignment of 32 bytes I think (but I'm not positive that's true once it bursts to the heap in an arc, it might be bigger at that point)

luciusmagn:
Hi, I am not sure if it may be any of any interest, but I wrapped a sled DB in a type-safe shell here:

<https://github.com/broadcom-codebois/backend/blob/rust/src/db.rs>

Perhaps it may be of useful to someone looking to do it now before this is merged :D (I understand this PR should provide similar functionality)

portstrom:
I'm hoping there will be less dependency on Serde, not more. I'm using another serialization library that is much faster than Bincode *and* produces much more compact data. Serde is obviously designed for JSON with all of its mistakes, and shouldn't be used for properly designed formats.

Apart from my first point, I don't see what typed trees would be good for. It seems like an attempt to imitate a feature from legacy databases without thinking whether this legacy feature is really something that has a place in a modern database. In an SQL database, data has to be arranged in tables, with one row type per table. Without SQL, data is free to be arranged in richer ways. I can for example make keys represent a hierarchy, such as */extensions/shop/products/42/categories*, and the data for this item would be a different type than */extensions/shop/products/42/pictures*. These items can even be queried together using one cursor, and stored in the same page for quickly loading all related items. The type depends on all levels in the hierarchy. In systems that use SQL databases, things have to be awkwardly shoehorned into a flat collection of tables, such as extensions for web applications prefixing all their table names with the name of the extension. A *hierarchy* of typed trees would be a better fit, but it still seems like an unnecessary feature, when I can simply put the hierarchy in my keys and not have to care to recursively create and remove trees in conjunction with creating and removing items that depend on being placed in a subtree of their own.

If someone needs a typed database API, I think the right way to go is to create another typed database library that uses Sled as its underlying store, thus keeping Sled unopinionated. That way we could see some interesting and innovative database APIs built on top of Sled.

spacejam:
That's a good point about heterogenous keyspaces, Fredrik. The main thing I
think about becoming nicer with typed trees is reducing repeated
deserialization costs for hot keys. I want to support people writing high
level databases, and I think it would be cool to write a few prototype
systems - sql, document, datalog, etc... and get a sense of common needs
before diving into adding this complexity to sled. I agree that by adding
something like this, it can add complexity that might not be as general
purpose as the other features. I'm going to close this issue for now, but
maybe add some related issues. I'm definitely interested in finding general
purpose ways of helping people to reduce deserialization costs for hot
keys, but right now I'm not sure if the approach in this PR is as general
purpose as sled's other features.

PR #807, opened 2019-09-07
Flush operation in failpoint tests

divergentdave:
This PR adds a separate Flush operation to the `test_tree_failpoints.rs` quickcheck tests. The idea for this was in one of the comments in the file. The reference data structure no longer holds a `(u16, bool)` indicating the last value, and whether it's certain or not, rather, it now it holds a `Vec<Option<u16>>` of possible values. The Set and Del operations add to the list of possible values, while the new Flush operation discards all but the last possible value.

I didn't discover any new bugs with this change. I'd welcome input on the probability distribution in `Op::arbitrary`. BTW, feel free to squash this PR, it was a real garden path, and I committed frequently along the way.

### DIFF
--- a/tests/tests/test_tree_failpoints.rs
+++ b/tests/tests/test_tree_failpoints.rs
@@ -170,7 +170,8 @@ fn run_tree_crashes_nicely(ops: Vec<Op>, flusher: bool) -> bool {
                 // make sure the tree value is in there
                 while let Some((ref_key, ref_expected)) = ref_iter.next() {
                     match ref_expected {
-                        Expected::Certain(_ref_value) => {
+                        Expected::Certain(None) => continue,
+                        Expected::Certain(Some(_ref_value)) => {
                             // tree MUST match reference if we have a certain reference
                             if t != ref_key {
                                 println!(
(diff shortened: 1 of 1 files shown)


spacejam:
Thanks for adding Flush as its own operation! 

It would be nice to include certainty as well, so that we can assert that the reference contains no items that we expect to be present after a flush but are no longer available, but this PR doesn't need it to merge, but it could be nice to have in the future

divergentdave:
I think what you describe is already handled; I'll do another pass on the comments and file a follow-up PR.

The case where the certainty bit is true maps to a Vec in the reference data structure that only has one value in it (either None or Some(_)). A false certainty bit maps to the Vec having multiple values (both None and Some(_)). The Flush operation turns all Vecs of two or more values into Vecs with one value, which maps to the certainty bit going from false to true.

divergentdave:
I'm now realizing this isn't sufficient, because crashes/restarts can happen in between the Set/Del operation and the Flush operation, so it isn't always correct to narrow down to the last possible value at Flush time. I think to address this, I'll need to add a crash counter, keep track of the "crash epoch" of the various expected values in the reference data structure, and then change Flush so that it doesn't throw away possible values if the entry was last touched before the most recent crash.

PR #815, opened 2019-09-08
Further improvements to failpoint test

divergentdave:
This is a followup to #807

* After iterating through the tree, check the rest of the reference, in case the tree is missing the last keys.
* Keep track of crashes, and keep the reference uncertain about operations that happened after a flush and before a crash.


spacejam:
Thanks for updating this!

PR #817, opened 2019-09-09
remove some code duplication

sitano:


sitano:
it seems like this failure 
```
The application panicked (crashed).
Message:  [quickcheck] TEST FAILED. Arguments: ([Set, Set, Set, FailPoint("buffer write post"), Del(0), Flush, Restart], false)
Location: /Users/vsts/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-0.9.0/src/tester.rs:178

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
test quickcheck_tree_with_failpoints ... FAILED

failures:

failures:
    quickcheck_tree_with_failpoints

test result: FAILED. 28 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '-p tests --test test_tree_failpoints'
```
is not related to this branch

spacejam:
Yeah, it's a random failure that got mined with the new failpoints code, and doesn't have anything to do with your pr

crates/sled/src/frag.rs
```diff
@@ -17,3 +20,16 @@ pub(crate) enum Frag {
     ParentMergeConfirm,
     ChildMergeCap,
 }
+
+impl Frag {
+    pub fn base(data: Data) -> Frag {
```
spacejam: could you rename this function `root` to reduce the chances of misuse over time? I want to be clear that it's not really valid for other purposes

divergentdave:
Yeah, that's my bad, #815 should fix that

crates/sled/src/frag.rs
sitano: @spacejam did you mean rename this function to `root`?

spacejam: yes, please. the reason is that the only kind of base that this creates a valid version of is for new tree root nodes. you could also assert that data.is_index() to give extra certainty

sitano: there is also a Leaf at crates/sled/src/meta.rs:35

sitano: @spacejam I have made a `base` private and exposed two methods `leaf` and `root` with assertions. Hope I get your point well. If not, don't hesitate to ping me back.

sitano:
another one
```
The application panicked (crashed).
Message:  [quickcheck] TEST FAILED. Arguments: ([Set, Set, FailPoint("blob blob write"), Set, Del(0), FailPoint("snap write crc"), Flush, Del(1)], false)
Location: C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\quickcheck-0.9.0\src\tester.rs:178

failures:
    quickcheck_tree_with_failpoints
```

spacejam:
Nice work! Thank you :]

Issue #679, opened 2019-06-12
move persisted config into the pagecache itself


spacejam:
this will allow us to get rid of the config external file

spacejam:
I've added a "PersistedConfig" type that is currently being added into the pagecache as a special page. This should be made to hold a few things that are currently being checked for in the written config file in crates/pagecache/src/config.rs in the verify_config_changes_ok function, specifically only `io_buf_size` and `use_compression`. However, we should also add the pagecache version to this, which will help to facilitate automated (not automatic, this will be manual but a 1-liner) upgrading

portstrom:
While you're at it, maybe you can drop the dependency on Serde at the same time. Serde is rather bulky, and obviously designed to parse text files of dubious correctness that need more code for handling edge-cases and formatting pretty error messages than for actually parsing the correct data.

sitano:
@portstrom do you mean just cutting everything related to the `serde` off the ground?

portstrom:
@sitano I would suggest making a very simple format similar to [Bincode](https://github.com/servo/bincode), but implementing the deserializer and serializer manually instead of having dependencies, as it's such a small feature.

spacejam:
Closing this due to the fact that when the config gets moved into the pagecache, it becomes a chicken-and-egg problem to actually detect which version it is. by keeping it in a separate file, it ensures we're able to properly do version detection going forwards.

PR #822, opened 2019-09-12
Persisted Config

sitano:
ref: https://github.com/spacejam/sled/issues/679


spacejam:
One issue that makes this a bit trickier, which I just remembered, is that there's a bit of a chicken and the egg problem here, where if the wrong config is used, it will never get to the step where it can read the last one, and it might just interpret the database file as garbage and delete it all. That's probably not a good outcome. I'm starting to think that we might want to keep the separate config file so we can detect changes before they accidentally delete anything.

It seems like most of this PR is cleaning other things up though, so I'm happy to merge in that part, but I think this config issue might actually cause data loss :/ 

Apologies for only realizing this after seeing the code

sitano:
:lol: how I missed it)

sitano:
@spacejam is that right that currently the only things that affect page compatability is the use of compression and db version? we could e.g. treat those system pages in a special way, don't compress them ever, or store and calculate a checksum in config not to lost a data. The checksum in config will solve the problem of invalid incoming configuration. What do you think?

spacejam:
I think the separate config file should be:
crc32 | u8 major LE | u8 minor LE | u8 compression bool 0 or 1 

even though we don't need a crc32 for 3 bytes of data, we're already using crc32 in a bunch of other places and it's better not to bring in any other dependencies or complexity. Over time we can add more fields in a grow-only manner.

write temp then mv to avoid tears

sitano:
Not sure we are talking about the same. I meant that we could use checksums for Persistent Config stored in the page to ensure that we read it right even with different incoming configuration from the client.

PR #828, opened 2019-09-18
extract sys_limits, improve few code moments, save basic data to p_config

sitano:


spacejam:
Looks good, thanks for these cleanups :)

sitano:
@spacejam have you seen that I left persisting some config data to persistent config? I thought that maybe its worth adding support to the migration tool to update veraion as well

PR #829, opened 2019-09-19
Remove io_buf_size and page_consolidation_threshold from ConfigBuilder

spacejam:


spacejam:
Over time, these may be dynamically tuned parameters based on whether users want a small database or a fast one.

spacejam:
@sitano this adds a separate cross-compilation script - note that it tests ios which only works on macos, or at least I've only had luck with it there

Issue #820, opened 2019-09-11
compare and swap refactor


spacejam:
- [ ] rename cas to compare_and_swap
- [ ] keep a cas method that just does a compiler error with a message to use compare_and_swap, which will have a different error signature (if it had the same signature we could do a deprecation, but because it changes, we should give people an error message that includes a short migration guide)
- [ ] add CompareAndSwapError that includes current and proposed-replacement values
- [ ] improve docs, especially around failures

sitano:
Trying to imagine a workaround for not being able to store Key and Values ref types in an Error Result struct due to the lack of generics in this enum and lifetime:
```
pub enum Error {
    ...
    /// A compare and swap failure
    CompareAndSwap {
        key: &[u8], <---------------------- require a lifetime which cant be passed here
        old: Option<&[u8]>,
        new: Option<crate::ivec::IVec>,
    },
}
```

spacejam:
closed by #832

PR #831, opened 2019-09-19
further code improvements related to basic db primitives naming

sitano:
- LogOffset makes code much much more readable and easier to understand
- because it now always divides the LSN or OFFSET by the SegmentSize

### DIFF
--- a/src/pagecache/mod.rs
+++ b/src/pagecache/mod.rs
@@ -59,19 +59,19 @@ pub use self::{
     segment::SegmentMode,
 };
 
-/// A segment id == Segment.LogOffset / Config.SegmentSize.
+/// The offset of a segment. This equals its LogOffset (or the offset of any
+/// item contained inside it) divided by the configured segment_size.
 pub type SegmentId = usize;
 
-/// An offset in the file storage sometimes called LogId (lid).
+/// A file offset in the database log.
 pub type LogOffset = u64;
 
 /// A pointer to an blob blob.
 pub type BlobPointer = Lsn;
 
-/// A logical sequence number of a log entry ==
-/// aligned log entry LogOffset ==
-/// Segment.LogOffset & ~SegmentSize
-/// Represent an Id of a log segment.
+/// A logical sequence number of a log entry.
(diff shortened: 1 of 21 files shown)


src/pagecache/mod.rs
```diff
@@ -59,16 +59,19 @@ pub use self::{
     segment::SegmentMode,
 };
 
-/// An offset for a storage file segment.
+/// A segment id == Segment.LogOffset / Config.SegmentSize.
```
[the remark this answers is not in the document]
sitano: I updated it as you asked. I think segment offset instead of Id is a little bit confusing because all offsets are used in relation to the database log. And the factual segment offset is expressed with LSN. Thus we are having 2 segment offsets by definitions.

src/pagecache/mod.rs
```diff
@@ -59,16 +59,19 @@ pub use self::{
...
+pub type LogOffset = u64;
 
 /// A pointer to an blob blob.
 pub type BlobPointer = Lsn;
 
-/// A logical sequence number.
+/// A logical sequence number of a log entry ==
+/// aligned log entry LogOffset ==
+/// Segment.LogOffset & ~SegmentSize
+/// Represent an Id of a log segment.
```
spacejam: "The logical sequence number of an item in the database log."

src/pagecache/mod.rs
```diff
@@ -59,16 +59,19 @@ pub use self::{
...
+/// A file offset in the database log.
+pub type LogOffset = u64;
 
 /// A pointer to an blob blob.
 pub type BlobPointer = Lsn;
 
-/// A logical sequence number.
+/// A logical sequence number of a log entry.
+/// A segment lsn equals to an aligned LogOffset that is
+/// Segment.LogOffset & ~SegmentSize.
```
spacejam: please no symbols or CamelCase names that don't match to real names in the code.

spacejam:
if you tackle #820, please work on it in a separate branch to keep different work in different PRs

spacejam:
Looks good! Thanks for the cleanup :)

I'll merge as soon as tests pass after merging in master

PR #832, opened 2019-09-20
new cas error and cas renamed to compare_and_swap

sitano:
this is impl of https://github.com/spacejam/sled/issues/820

@spacejam do you have an idea of how to raise a compile error only when someone calls `cas`?

### DIFF
--- a/src/result.rs
+++ b/src/result.rs
@@ -5,9 +5,9 @@ use std::{
     io,
 };
 
-use crate::pagecache::{
-    PagePtr, DiskPtr
-};
+use crate::pagecache::{DiskPtr, PagePtr};
+
+use crate::ivec::IVec;
 
 /// The top-level result type for dealing with
 /// the `PageCache`.
@@ -42,6 +42,11 @@ pub enum Error {
     #[doc(hidden)]
     #[cfg(feature = "failpoints")]
     FailPoint,
+    /// Compare and swap failure.
+    CompareAndSwap {
+        /// Current value if present.
+        cur: Option<IVec>,
+    },
 }
 
 impl Clone for Error {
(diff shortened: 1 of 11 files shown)


src/result.rs
```diff
@@ -42,22 +42,26 @@ pub enum Error {
     #[doc(hidden)]
     #[cfg(feature = "failpoints")]
     FailPoint,
+    /// Compare and swap failure.
+    CompareAndSwap {
+        /// Current value if present.
+        cur: Option<IVec>,
+    },
```
spacejam: CompareAndSwap should not be a member of this Error enum, because it represents a high-level normal failure and not an operational critical error. In the past, these were mixed, and it led to a lot of bugs because it basically makes exhaustive pattern matching useless on Error types. It's better to use nested result types when dealing with multiple levels of concerns, even though it feels more heavy weight. It pushes the real complexity at play into awareness of the user.

sitano:
Oops

src/tree.rs
```diff
@@ -407,28 +407,41 @@ impl Tree {
...
     /// assert_eq!(t.get(&[1]), Ok(None));
     /// ```
-    pub fn cas<K, OV, NV>(
+    pub fn compare_and_swap<K, OV, NV>(
         &self,
         key: K,
         old: Option<OV>,
         new: Option<NV>,
-    ) -> Result<std::result::Result<(), Option<IVec>>>
+    ) -> Result<()>
```
spacejam: This should return a nested result because the results represent different kinds of failures. In the past they were mixed and it led to a lot of bugs. When I switched over to using nested results, the refactor caught several bugs that had not been caught in testing yet as well. That experience led me to believe that very different concerns should not share the same error enums, and it's better to use nesting instead to make the failure handling more clear.

```
) -> Result<std::result::Result<(), CompareAndSwapError>>
```

spacejam: Also feel free to make a public type alias in src/result.rs and re-export it from src/lib.rs:
```
pub type CompareAndSwapResult = std::result::Result<(), CompareAndSwapError>;
```

PR #833, opened 2019-09-20
Tree::cas documentation clarification

jhwgh1968:
The documentation currently says that if the ``Tree::cas()`` operation succeeds, the value returned is a ``Result<(), Option<IVec>>`` -- without really explaining why, or what the inner error value actually is.

This PR adds a doc test to provide the clarity I was missing.


spacejam:
Thanks :) This has been a sore spot for a while, but is about to get a nice facelift in other ways too.

PR #836, opened 2019-09-22
[wip] fully-compressed Arc-Node

spacejam:


spacejam:
The Arc<Node> work has a bit more to go, but I'm going to merge this now to reduce things in-flight, and circle-back to the work in a future PR.

PR #838, opened 2019-09-24
tree::cas -> tree::compare_and_swap, CAS_OK, CompareAndSwapResult/Error

sitano:
this is impl of https://github.com/spacejam/sled/issues/820

`sled_native` now compiles, not sure it works tho.

@spacejam do you have an idea of how to raise a compile error only when someone calls `cas`?


Kerollmops:
You can see that we already use the deprecate attribute here

https://github.com/spacejam/sled/blob/d92239addcc853859648bfcc60885f4b7f6e587d/src/tree.rs#L100

sitano:
@Kerollmops yes, but it does not produce compilation error which I hoped for to show only when there is a real call of a deprecated method

Kerollmops:
Ho my bad, I think you can just remove the function, so :)

sitano:
I just didn't understand how to implement this: 

> keep a cas method that just does a compiler error with a message to use compare_and_swap

Kerollmops:
I think @spacejam means a compiler warning like the other deprecated methods

src/tree.rs
```diff
@@ -2009,3 +2057,34 @@ impl Debug for Tree {
...
+    fn from(error: Error) -> Self {
+        Err(error)
+    }
+}
+
+/// Compare and swap error.
+#[derive(Debug, Clone, PartialEq)]
+pub struct CompareAndSwapError {
+    /// Current value.
+    pub cur: Option<IVec>,
```
spacejam: `cur` -> `current` to be more explicit

I'll add another field that returns the proposed new value that failed to be applied, but don't worry about this for this PR

src/tree.rs
```diff
@@ -494,12 +510,32 @@ impl Tree {
...
+        OV: AsRef<[u8]>,
+        IVec: From<NV>,
+    {
+        match self.compare_and_swap(key, old, new) {
+            Ok(Ok(())) => Ok(Ok(())),
+            Ok(Err(CompareAndSwapError { cur })) => Ok(Err(cur)),
+            Err(e) => Err(e),
+        }
+    }
+
```
spacejam: nice adapter that prevents breakage!

src/tree.rs
```diff
@@ -494,12 +510,32 @@ impl Tree {
                     res.complete(event);
                 }
 
-                return Ok(Ok(()));
+                return CAS_OK;
```
spacejam: Please remove `CAS_OK` because it's confusing until they look at docs. Everyone knows `Ok(Ok(()))` without having to look at docs, so it's less effort to use.

spacejam:
Sorry it took me so long to review this, I just got back from 2 weeks of travel. Thank you for implementing this! I'll merge this right away after removing CAS_OK in favor of raw `Ok(Ok(()))` and making `cur` more explicit as `current`! Nice work :)

sitano:
@spacejam glad you are back! hope your travel went great!

sitano:
@spacejam I added `proposed` value to the cas err. If you would like to have it diff ping.

spacejam:
Thanks :) I'll merge this right after tests go green after merging in master.

PR #841, opened 2019-10-01
tree node optimizations

spacejam:


spacejam:
This PR adds some cool classic B-Tree node optimizations:
* cleans up prefix encoding from being per-key to being per-node- which makes the code simpler and usually makes the compression better
* adds suffix truncation to make indexes more tightly packed
* adds interpolation search to speed up the first 2 rounds of index searching

Probably index nodes can now be made a lot wider now, to reduce tree depth. That will come later after more measurement and tuning.

PR #847, opened 2019-10-05
Big configuration rework

spacejam:
* Remove ConfigBuilder
* Add Config::open
* Persist config settings using a simple text format that does not rely on serde
* Clean up test_log.rs


spacejam:
Closes #711

PR #858, opened 2019-10-11
Add Transactional impl over 3-tuple of &Tree.

thedodd:
Long-term, we will probably want a macro for more easily generating
these impls. Let's discuss.

@spacejam I know you mentioned wanting to put a macro together for this. I figured I would go ahead and open this PR now, we can chat about details on the macro, and this will also unblock me so I can start using this ASAP.


spacejam:
Closing this because I believe #859 resolves the underlying issue

PR #859, opened 2019-10-11
Macro, impl Transactional on more tuples of trees

divergentdave:
This is parallel to #858, but uses a macro to implement Transactional on tuples of two to ten trees. The design of the macro was largely driven by the syntax for indexing into tuples, which requires literals. As such, the main macro takes sequential numbers as arguments. The `repeat_type!` macro uses recursion to turn the list of numbers into a list of a type, repeated that many times. This is used to define the tuple for which the trait is implemented, as well as the `View` associated type.

### DIFF
--- a/src/tx.rs
+++ b/src/tx.rs
@@ -329,27 +329,70 @@ impl Transactional for &Tree {
     }
 }
 
-impl Transactional for (&Tree, &Tree) {
-    type View = (TransactionalTree, TransactionalTree);
-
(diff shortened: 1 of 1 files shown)


spacejam:
Looks great, thanks for implementing this! I'll add an entry to the changelog about this being implemented too, since it adds new functionality.

thedodd:
Thanks @divergentdave! Solid macro `:)`

Issue #571, opened 2019-03-06
io_uring support


spacejam:
io_uring is an incredibly exciting new linux interface that in some situations is able to outperform SPDK. 

* supports mapping kernel memory for IO buffers into userspace
* supports both buffered and direct IO, so we can make progress on some parts of this independently of #545 possibly
* kernel-side submissions for IO with 0 syscalls

c example usage with O_DIRECT:
http://git.kernel.dk/cgit/fio/plain/t/io_uring.c

thedrow:
We can provide bindings to [liburing](http://git.kernel.dk/cgit/liburing/) which enables us to use it correctly.
I wonder how well it can play with tokio and the upcoming stabilized async/await mechanism.

thedrow:
There is a Rust wrapper to the `io_uring` mechanism that integrates with tokio [here](https://github.com/stbuehler/rust-io-uring).
It is obviously very much alpha.

I hope I helped.

sitano:
> kernel-side submissions for IO with 0 syscalls
It's a great option but having rare submission events would put kernel thread to sleep. I don't know what is his timeout, probably have to check kernel sources, but then it requires a syscall with AWAKE flag set on.
> liburing
there is https://github.com/polachok/liburing-sys so we can just derive

quininer:
I just implemented [a pure Rust io-uring library](https://github.com/quininer/linux-io-uring), I believe it will help.

PR #872, opened 2019-10-25
[WIP] simple io_uring implementation

sitano:
At least safe 1 syscall on write path. Don't wait for fsync.

Implementation:
- separate write uring queue for writes
- writes usually require 2 SQE per write (write+fsync)
- fsyncs are issued with DRAIN and not waited
- URing uses preallocated iovecs structs and keep Arcs for used bufs
- when submit queue exhausts write() will wait for minimum empty slots
- uring is used via my custom binding to liburing similar to https://github.com/sitano/liburing

URing.write_at(buf, offset, fsync) implementation is as follows:
- take lock
- drain completion queue to retain free slots
- reserve required number of SQEs
- build write
- build fsync if needed
- submit

ISSUE: https://github.com/spacejam/sled/issues/571

TODO:
- configuration
- sqpoll support
- CI for io_uring on linux

FUTURE:
- fixed buffers + buffer pool for write path
- SQPOLL?


liburing/build.rs
```diff
@@ -0,0 +1,45 @@
...
+    let out_dir = env::var("OUT_DIR").unwrap();
+
+    Command::new("cp")
+        .arg("-r")
+        .arg("lib")
+        .arg(out_dir.clone())
+        .status()
+        .expect("copy liburing to out_dir");
+    Command::new("make")
+        .arg("liburing.a")
```
quininer: I think we can't statically link liburing because of LGPL, maybe you can consider using [linux-io-uring](https://github.com/quininer/linux-io-uring).

quininer: ~~I think we can't statically link liburing because of LGPL, maybe you can consider using [linux-io-uring](https://github.com/quininer/linux-io-uring).~~

> a) Accompany the work with the complete corresponding machine-readable source code for the Library including whatever changes were used in the work (which must be distributed under Sections 1 and 2 above); and, if the work is an executable linked with the Library, with the complete machine-readable "work that uses the Library", as object code and/or source code, so that the user can modify the Library and then relink to produce a modified executable containing the modified Library. (It is understood that the user who changes the contents of definitions files in the Library will not necessarily be able to recompile the application to use the modified definitions.)

I re-read the LGPL, it not infect LGPL to other libraries, but still has restrictions on final executable.

sitano: I don't think it's worth to replace original liburing with pure Rust implementation. liburing is kinda standard way and supported library for working with io_uring. work is still in progress there. there is no sense to support another library porting code in par.

quininer: The developer of `rust-io-uring` has answered a similar [question](https://github.com/stbuehler/rust-io-uring/issues/1), and my point of view is similar to it. I think Rust is better suited to implement this interface than C, and can provide a safer and simpler API.

But now I think using `liburing` is good for `sled`, because `liburing` has a `libaio` [compatibility layer](https://github.com/axboe/liburing/tree/libaio-wrapper). This will provide `sled` with the old version of linux compatibility for free.

sitano: Thank you for the link. Generally, I think pure Rust implementations are a good idea. In this certain case, it makes no difference, besides that the interface is more natural. I've read mentioned Issue #1, and all that author has to say is his subjective perception of liburing.

PR #886, opened 2019-11-14
WIP: io_uring: minor updates

sitano:
continuing https://github.com/spacejam/sled/pull/877.

- use IOSQE_FIXED_FILE for writing registered fd
- fix converting io_uring errors to os codes

### DIFF
--- a/src/pagecache/io_uring.rs
+++ b/src/pagecache/io_uring.rs
@@ -64,7 +64,7 @@ impl<T> Uring<T> {
                 flags,
             );
             if ret < 0 {
-                return Err(Error::Io(io::Error::from_raw_os_error(ret)));
+                return Err(Error::Io(io::Error::from_raw_os_error(-ret)));
             }
 
             s.assume_init()
@@ -85,7 +85,7 @@ impl<T> Uring<T> {
                 1,
             );
             if ret < 0 {
-                return Err(Error::Io(io::Error::from_raw_os_error(ret)));
+                return Err(Error::Io(io::Error::from_raw_os_error(-ret)));
             }
         };
 
@@ -112,7 +112,7 @@ impl<T> Uring<T> {
                 return Ok(());
             }
             if ret < 0 {
-                return Err(Error::Io(io::Error::from_raw_os_error(ret)));
+                return Err(Error::Io(io::Error::from_raw_os_error(-ret)));
             }
             let i = usize::try_from((*cqe).user_data).unwrap();
             if i < self.free_slots.capacity() {
@@ -152,7 +152,7 @@ impl<T> Uring<T> {
                     libc::c_uint::try_from(reserve).unwrap(),
                 );
                 if ret < 0 {
-                    return Err(Error::Io(io::Error::from_raw_os_error(ret)));
+                    return Err(Error::Io(io::Error::from_raw_os_error(-ret)));
                 }
                 self.drain_cqe()?;
                 if self.free_slots.len() < reserve {
@@ -186,6 +186,7 @@ impl<T> Uring<T> {
                     1,
                     off_t::try_from(offset).unwrap(),
                 );
+                io_uring_sqe_set_flags(sqe, IOSQE_FIXED_FILE);
                 (*sqe).user_data = u64::try_from(i).unwrap();
             }
 
@@ -213,7 +214,7 @@ impl<T> Uring<T> {
             // 5. submit
             let ret = io_uring_submit(&mut self.ring);
             if ret < 0 {
-                return Err(Error::Io(io::Error::from_raw_os_error(ret)));
+                return Err(Error::Io(io::Error::from_raw_os_error(-ret)));
             }
         };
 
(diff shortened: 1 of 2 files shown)


sitano:
@spacejam if we want to use preregistered buffers (IORING_OP_WRITE_FIXED https://github.com/axboe/liburing/blob/master/man/io_uring_enter.2#L165) we need to preallocate pool of buffers of SQ.len / 2 length which for the segments will be 8 megs each * 16 by default (we need to make it smaller) = 128 megs of waste just for making segments writes. Not sure it's worth it. The min recommended number of buffers must be at least 1 / flushes / sec (for 200 ms flushes = at least 5 buffers). Not sure we want it now.

Do you still see ENOMEM and wait sqes errors?

src/pagecache/io_uring.rs
```diff
@@ -186,6 +186,7 @@ impl<T> Uring<T> {
                     1,
                     off_t::try_from(offset).unwrap(),
                 );
+                io_uring_sqe_set_flags(sqe, IOSQE_FIXED_FILE);
```
quininer: IOSQE_FIXED_FILE will use the index of registered fd list instead of fd itself.

sitano: interesting, why does it even worked

spacejam:
when running `cargo test --release --features=testing,pretty_backtrace,io_uring` I get "wait for free sqes" errors when running tests that spin up lots of processes in short order

spacejam:
@sitano would you prefer that I merged this in to keep it moving, since this already includes some fixes? I'm still getting:
```
benchmarks/stress2 [0647baf ●]
λ cargo run --release --features=io_uring,logging
   Compiling sled v0.29.2 (/home/t/src/sled)
   Compiling stress2 v0.1.0 (/home/t/src/sled/benchmarks/stress2)
    Finished release [optimized + debuginfo] target(s) in 27.42s
     Running `target/release/stress2`
did 2507309 ops
did 2550314 ops
did 2530137 ops
ERROR sled-io                   iobuf      hit error while writing iobuf with lsn 243269632: Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })
ERROR log flusher               flusher    failed to flush from periodic flush thread: IO error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
Oh noez! Panic! 💥
Message:  called `Result::unwrap()` on an `Err` value: Io(Custom { kind: InvalidInput, error: "Os { code: 22, kind: InvalidInput, message: \"Invalid argument\" }" })
Location: src/libcore/result.rs:1165

Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                         (11 post panic frames hidden)
11: stress2::run::h414007a4928f4bc4
    at src/main.rs:0
12: __rust_maybe_catch_panic
    at src/libpanic_abort/lib.rs:28
[1]    63539 abort (core dumped)
```

and

```
etable::test_split_fanout ... ok
test subscription::basic_subscription ... ok
test stack::basic_functionality ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/test_crash_recovery-36cd28d3995766a7

running 4 tests
test test_crash_batches_no_runtime_snapshot ... ignored
test test_crash_batches_with_runtime_snapshot ... ignored
ERROR sled-io              iobuf      hit error while writing iobuf with lsn 12288: Io(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: Io(Custom { kind: InvalidInput, error: "Os { code: 22, kind: InvalidInput, message: \"Invalid argument\" }" })
```

sitano:
@spacejam merge. it contains important fixes, but they don't fix all of them. Issues that are left to resolve:
- getting `wait on sqes`, which is nonsense for `env RUST_BACKTRACE=1 cargo test --features=testing,io_uring test_crash_recovery_with_runtime_snapshot` alone
- having EINVAL which is nonsense too for newly committed fsync and writes. mainly appear when there are fsyncs in queue processing.  I need to look into kernel source to understand this. maybe add some stap, ebpf, kprintf.
- ENOMEM. this is nonsense as well. my current blindly generated idea is that there are lot of parallel Idon'tKnow allocations - aka buffers registration in example, due to the parallel tests executions or somewhat else. I think that maybe it must be the easiest. 
Still going to look into them. But they require much more time and attention.

sitano:
@spacejam I am sorry it takes so much time and previous branch did not work as expected.

spacejam:
@sitano you have nothing to apologize for! You're doing really good work on this, especially considering how you're one of the first people on earth to be diving into this new kernel API that is going to become more and more important over time I think. That's a pretty cool skill to be gaining, and all of the learning mistakes are sealed off of the main users with the feature flag so it's a safe place to learn. I'm really grateful that this is happening because it also lets me learn about a few sharp edges without spending a lot of time diving into it myself.

Issue #530, opened 2019-02-06
blob_io.rs failpoints 


spacejam:
blob_io needs more thorough failpoints

spacejam:
There are 2 small pieces of logic need to be added here:
1. a similar way to trigger a failpoint, as is done in iobuf.rs in the pagecache crate
2. adding the names of the new failpoints to the test_tree_with_failpoints.rs list of failpoints in the Arbitrary trait for generating random failpoints to trigger during testing

spacejam:
Closed for now in #888, thanks @peter-wilkins! For reader failpoints (#666) we will need to think about a different testing strategy, as the assumptions around failpoints in the testing model assume all previous writes are able to be read after an fsync, before new failpoints are added. reader failpoints turn what is assumed to be stable into something that may not be readable, so they need to apply more directly to never panicking, but always returning a Result of some kind, whether it's a success or a failure.

Issue #666, opened 2019-05-24
failpoints on disk reads


spacejam:
Thank you for understanding :)

spacejam:
This needs to be performed in a different way than how our current failpoints are being handled, as when a failpoint is triggered, all writes from that point onward until the next fsync are considered to be stable. For this work, we need to relax the assumption, and only ensure that no operation ever panics. Errors on disk reads should always return an Err, rather than panicking, so this can actually be a much simpler model than the existing failpoint test.

PR #888, opened 2019-11-15
WIP failpoints in blob_io.rs

peter-wilkins:
added fail points for all io operations (except in gc_blobs)

test_tree_failpoints test did not trigger any fail points in gc_blobs
so I removed them

issue: https://github.com/spacejam/sled/issues/530

### DIFF
--- a/src/pagecache/blob_io.rs
+++ b/src/pagecache/blob_io.rs
@@ -1,11 +1,22 @@
 use crate::pagecache::*;
 use crate::*;
 
+macro_rules! io_fail {
+    ($config:expr, $e:expr) => {
+        #[cfg(feature = "failpoints")]
+        fail_point!($e, |_| {
+            $config.set_global_error(Error::FailPoint);
+            Err(Error::FailPoint)
+        });
+    };
+}
+
 pub(crate) fn read_blob(
     blob_ptr: Lsn,
     config: &Config,
 ) -> Result<(MessageKind, Vec<u8>)> {
     let path = config.blob_path(blob_ptr);
+    io_fail!(config, "read_blob open file");
     let f_res = std::fs::OpenOptions::new().read(true).open(&path);
 
     if let Err(e) = &f_res {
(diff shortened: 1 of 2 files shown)


peter-wilkins:
These new fail-points cause test to fail on my machine (ubuntu). 
Wanted to check in and see if I've done it right.
I would like to investigate the failures if they are legitimate but a few pointer would probably help quite a lot. Also how do you re-run a test with he seed? 
Great work btw.

spacejam:
Hey @peter-wilkins, thanks for adding these! 

These results are causing me to realize that failpoints on the readpath (#666) will need a different testing strategy.

The reason only the read-path ones are failing is because the general assumption of the simplified test model is that anything that has been written and fsynced will be readable in the future, and it is moved to a structure that represents expected values that will be read later. When those future reads fail due to the new failpoints, it causes the model's expectation to be violated.

The test begins to flag certain progress as being uncertain as soon as a failpoint is activated, so I think the best thing to do for now is to move these read failures to be write failures that will play better with the existing model's correctness assumptions. So, for each of the read_blob failpoints you added, either make sure that a write-side failpoint is already being exercised, or add one if not.

So, remove all read_blob failpoints for now, remove write_blob write all, split the 3 writes that are currently chained with and_then in the write_blob function, and possibly fail them all with specific failpoints for:
* write_blob write crc
* write_blob write kind_byte
* write_blob write buf

opening of the file is already covered by `write_blob open file` which will skip its creation. 

There is definitely also room to add another model that checks our assumptions about things that can be read, but your work here has shown that we would need a different kind of model to reason about that, and it's not yet clear to me what bugs would be detectable from that which are not detectable with the current write-side flagging of uncertainty. In the end, the model needs to be kept simple, but read failures will happen after things are correctly written, we just need to figure out if it's ever OK to keep running the database when that happens, which I'm a bit skeptical about currently.

spacejam:
Nice work, thank you, @peter-wilkins :)

Issue #898, opened 2019-12-01
remove serde


spacejam:
on my main dev machine, serde takes up 16 of the 29 seconds required to do a release build of sled. human optimizations are the most important for long-term system health, since they encourage all other optimizations. sled doesn't need very complex serialization, so this will be fairly easy to replace.

pmuens:
Sounds like a good plan :+1:

> sled doesn't need very complex serialization, so this will be fairly easy to replace.

Any idea / proposal what `serde` can be replaced with?

spacejam:
@pmuens I replaced it with a pretty simple implementation in #902 that just uses some simple rust traits to write very little code, and it still performs on-par with serde even though now it takes half the time to compile the project!

pmuens:
@spacejam that's awesome!

I was wondering if there might be a need for a more lightweight serde-like crate which only provides the most minimal boilerplate to get basic serialization working.

I really enjoy the minimalism of `[serialization.rs](https://github.com/spacejam/sled/blob/e6a90afe77f8ed2f3ef5c79fa0a9cebea2c4c7d7/src/serialization.rs)`. Do you think there's room to make this more generic? Or do we end up with a reimplementation of serde once we'd go down that path?

spacejam:
@pmuens yeah! @Geal has a nice crate, [cookie factory](https://docs.rs/cookie-factory) that is a pretty clean and generic implementation of a simple serializer. It should be noted though that serde is really good due to the human convenience factor, and bincode used with serde is fast enough for many workloads. Once you go down the hand-written (even library-assisted) path, it's important to start spending a lot more effort on testing as well. That's why like half of [the new serialization module](https://github.com/spacejam/sled/blob/master/src/serialization.rs) in sled is just quickcheck tests. 

There are a few key reasons why I eventually chose to remove serde after 4 years of it being a great choice:

1. Fresh compile times actually make up a big part of my experienced friction when working with this project. I have the philosophy that human optimizations are the most important ones for building fast reliable software over long time periods. Anything that makes the project less exhilarating will ultimately lead to slower and buggier code, because the humans will be less psyched about pushing it forward. It needs to feel like a game. That's why I keep playing the "build sled" video game after 4 years, I've kept it fun for myself. Shaving off 50% of the compile time really makes trying out changes significantly more exhilarating for me. (I should probably be putting more effort into gamifying "release 1.0" as well though xD)
[...]

PR #902, opened 2019-12-10
no serde

spacejam:


spacejam:
Closes #898. This cuts sled compilation time in half, and has performance that is basically the same. This can be improved more easily over time by making these more SIMD-friendly.

spacejam:
Closes #898. This cuts sled compilation time in half, and has performance that is basically the same. This can be improved more easily over time by making these more SIMD-friendly.

PR #910, opened 2019-12-17
Fix bug with threadpool workers. Closes #909.

### LINKED ISSUE #909: Use 200% CPU when idle (0.30).
1. sled 0.30 (only happens with 0.30, 0.29 no problem)
2. rustc (1.39)
3. operating system (macosx)

Minimal reproducable code:

```
use sled::{Config, Db};
use std::thread::sleep;
use std::time::Duration;

fn main() {
    let db = Db::open("./data").unwrap();
    sleep(Duration::from_secs(9999999999999));
}
```

Profile result with XCode instruments:
image

spacejam:


spacejam:
closes #909

PR #915, opened 2019-12-25
Remove Pio trait, use bare functions

spacejam:


spacejam:
There is no reason for this trait to exist. And its name is bad.

PR #917, opened 2019-12-27
Add test for CPU usage with no DB operations

### LINKED ISSUE #911: measure cpu instructions for the null workload in CI test
#909 was a silly regression where when a database was idle, it was using a ton of CPU while spinning in a tight loop. In order to catch similar bugs in the future before they end up in a release, we can use cachegrind to count instructions of a "null workload" that just sleeps for 10 seconds, and errors out if it goes above a certain tripwire threshold.

steps:

- [ ] find out how to run cachegrind in the azure pipeline CI ([config](https://github.com/spacejam/sled/blob/master/azure-pipelines.yml))
- [ ] create a test binary that just sleeps after opening a DB:
```rust
fn main() {
  let _ = sled::Db::open("sleeper").unwrap();
  std::thread::sleep(std::time::Duration::from_secs(10));
}
```

divergentdave:
Fixes #911. This adds a new test executable that forks, opens a database, waits ten seconds, and drops it. The child process checks its own CPU usage using `libc`, and exits with an error code if it used more than one second. I tested it by making the flusher spawn an infinite loop in a new thread, and here's what the error messages look like.

```
running 1 test
thread '<unnamed>' panicked at 'Database used too much CPU during a quiescent workload. User: 10.262479s, system: 0.015997s (wall clock: 10.262207544s)', tests/test_quiescent.rs:36:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
test test_quiescent_cpu_time ... FAILED

failures:

---- test_quiescent_cpu_time stdout ----
thread 'test_quiescent_cpu_time' panicked at 'child exited abnormally', tests/test_quiescent.rs:61:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.


failures:
    test_quiescent_cpu_time

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -36,3 +36,11 @@ pub fn setup_logger() {
 
     let _r = builder.try_init();
 }
+
+#[allow(dead_code)]
+pub fn cleanup(dir: &str) {
+    let dir = std::path::Path::new(dir);
+    if dir.exists() {
+        std::fs::remove_dir_all(dir).unwrap();
+    }
+}
--- a/tests/test_crash_recovery.rs
+++ b/tests/test_crash_recovery.rs
@@ -2,15 +2,15 @@
 
 mod common;
 
-use std::fs;
 use std::mem::size_of;
-use std::path::Path;
 use std::thread;
 use std::time::Duration;
 
 use rand::Rng;
 
 use sled::Config;
+
+use common::cleanup;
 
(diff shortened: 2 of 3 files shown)


spacejam:
Nice work, @divergentdave! I really appreciate all of your help to make sled more reliable over time through this great testing work :]

Issue #897, opened 2019-12-01
optimistic concurrency control & correct batch implementation


spacejam:
putting this in one issue because these issues are two sides of the same coin, and it helps me keep notes in one place

concurrent transactions occurring on non-overlapping sets of trees can cause non-atomic recovery of transactions. we need to make transaction batching fully atomic by ditching the "plug" approach and just support multi-item logging (this also will reduce contention on the log).

- [ ] add relative file offset to blob pointers
- [ ] add batch log message that includes the expected batch length
- [ ] add reserve_batch method to Log
- [ ] recovery can be parallelized before the last batch header (after can also be done but is more complex and not likely to be as useful)
- [ ] switch recovery over to new batch style
- [ ] implement cicada-like page concurrency control
- [ ] add page view local write cache and read ts maps

PR #920, opened 2019-12-28
[WIP] Add batches to failpoint quickcheck test

divergentdave:
This adds batch operations to `test_tree_failpoints`. I'm aware that batches will be overhauled in light of #897, so it's too early to merge this just yet.

I've noticed two classes of failures from running this:

* When the batch is applied, a FailPoint error bubbles up inside `<Reservation as Drop>::drop` and gets unwrapped. Should this use `global_error` instead of unwrapping?
* Some test runs will report `failed to make_stable after 30 seconds.` I haven't dug into this much yet.


divergentdave:
Closing in favor of #1041

PR #921, opened 2019-12-28
Use seek_read/seek_write on Windows

divergentdave:
On Windows, the standard library provides [`seek_read`](https://doc.rust-lang.org/std/os/windows/fs/trait.FileExt.html#tymethod.seek_read) and [`seek_write`](https://doc.rust-lang.org/std/os/windows/fs/trait.FileExt.html#tymethod.seek_write), which are similar to `read_at` and `write_at`, except they also update the file's cursor before they return. (It appears that no part of sled relies on file cursor positions) This PR adds another implementation of `pread_exact` and `pwrite_all`, using these two methods, along with loops to finish partially read or written buffers. This change should hopefully improve performance on Windows by removing a global lock and allowing I/O operations to run in parallel with each other.


divergentdave:
The test failure is unrelated, and I can't easily replicate it locally

spacejam:
Looks good! Thanks!

PR #925, opened 2020-01-04
fix warning: `description` is deprecated since 1.41.0: use the Display impl or to_string() 

exfly:
…string()


spacejam:
Thanks!

PR #929, opened 2020-01-14
Link to import/export docs where README.md mentions manual migrations

JJJollyjim:
The import/export names aren't super discoverable in and of themselves given the description in the readme.


spacejam:
Nice addition, thanks!

PR #934, opened 2020-01-20
release linearizing mutexes before notifying condvars

spacejam:
This reduces contention by dropping the associated mutex before notifying its condvar. It is safe because anything that is waiting is waiting on specific state to materialize, which is guaranteed to happen before the condvar is pinged.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -698,6 +698,7 @@ impl Config {
         if ge.is_null() {
             Ok(())
         } else {
+            #[cold]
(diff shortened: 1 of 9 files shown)


spacejam:
This drops 12-core lexicographically-sequential writing of 100m small keys & values from 102 seconds to 90 :) Nice impact for high-contention workloads.

spacejam:
Also this drops log reservation retry-related contention by 20%

PR #942, opened 2020-01-29
Add new Medium IVec variant for data between 22 and 2048 bytes long.

spacejam:
Cloning this is faster than bumping an Arc.


spacejam:
Needs benchmarks, but based on some measurements of Arc cloning, the break-even point is around 2k. Need to re-run benchmarks accounting for possible allocation variance.

spacejam:
benchmarking showed this to be way slower than master for keys and values of 1024 bytes. It's possible there's a better threshold, but for now I'm going to just close this since it also causes memory to blow up more.

PR #943, opened 2020-01-30
use varints for message headers

spacejam:
this cuts headers down to 8 bytes sometimes. the first 5 bytes are never variable: the crc32 and the message kind byte. crc32 will often have all bits exercised, and if this was a varint it would very often use 5 bytes instead of 4. this is expected to have a similar impact as the nice gains we saw from using varints in the serialization of tree nodes. benchmarks incoming.

this also has the nice refactor of performing serialization AFTER successfully reserving the log slot, which currently might not impact anything, but it is now refactored such that it will be easy to perform the actual serialization in a conditional way later on after we know if our associated lock free operation on the tree succeeded, reducing the contention by performing less work between the in-memory read, the reservation in the middle, and then the in-memory atomic CAS.


spacejam:
Just ran a 100m small insert workload, 4 byte keys and 0 byte values. compared with master:
this: 80 seconds, 1.25m writes/s, 2.1gb written total, 136m storage directory after inserts
master: 92 seconds, 1.07m writes/s, 3.67db written total, 137m storage directory after inserts

this reduces bytes written by 42%, while increasing throughput by 17%. Very happy with these results after spending so long on this refactor :)

spacejam:
the reason it only uses 137mb for 400mb of logical data is mostly because of prefix encoding on the tree nodes

PR #946, opened 2020-01-30
use rio for io_uring workloads

spacejam:


spacejam:
cc @sitano 

one thing I found while using this is that it's possible to start throwing errno 22 if multiple threads try to use the uring without synchronization due to a bug in rio that is only triggered when events are submitted from another thread while completions are pending.

crusty-dave:
Use of rio is a show-stopper as it requires GPL 3 licence.

azastrael:
> Use of rio is a show-stopper as it requires GPL 3 licence.

Indeed. Doesn't it also mean that this repository needs to be relicensed under the GPL too, by the way?

crusty-dave:
@azastrael I am not sure, however, it means that lobbying my company to support sled is now not possible.  Our legal dept. doesn't typically approve requests to use open software under any GPL or MPL license without a lot more gnashing of teeth.  Not worth it, so I will now be completely abandoning any thoughts of using this promising crate.  I was just running into performance issues on Windows, which might be resolved by 0.31, but it is not even worth testing it now.

Licenser:
The inclusion of rio does not necessarily mean this crate is under GPL. As far as I understand, spacejam as the license holder could include it / relicense it for use with sled with any license he pleases if they want to. It just makes it a bit more complicated. That said IANAL. :)

spacejam does offer `rio` under the MIT/Apache 2.0 (https://github.com/spacejam/rio/blob/ce38304cbdd6afd99e874087ef8f53cdf8eb65bb/Cargo.toml#L6) for sponsors so @crusty-dave if you want to use it in a commercial setting petition your employer to chip in a few bucks?

azastrael:
@Licenser Sadly, this seems to mean that `rio` cannot accept contributions (otherwise, spacejam wouldn't own 100% of the code anymore). Also, in the mean time, `sled` is getting `rio` from the crates.io registry on which it is licensed under the `GPL-3.0`. To me, it seems like `sled` is depending on a library licensed under the GPL at the moment and should be relicensed under it too.

There are solutions, though:
1) Relicense `rio` for everyone; but I completely understand why this would be an issue.
2) Relicense `sled` to follow a model similar to `rio`; but this would mean asking every past contributors to accept this.
3) Make `rio` an optional feature (eventually with 3 options: disabled, enabled with the GPL version and enabled with the sponsors version).

Licenser:
> EDIT: (I just realized what your username is 😊)

Funny enough it is pure coincidence!

crusty-dave:
Apparently I missed that rio is an optional feature, so I am not sure what this means in terms of sled licensing?  If we don't use the u_ring feature (that pulls in rio), does sled remain free of gpl license restrictions?

```
[features]
...
io_uring = ["rio"]

[dependencies]
...
rio = { version = "0.9.2", optional = true }
```

Is there a guarantee that it will remain optional and that no required features will fall under a gpl license?

azastrael:
> Apparently I missed that rio is an optional feature, so I am not sure what this means in terms of sled licensing? If we don't use the u_ring feature (that pulls in rio), does sled remain free of gpl license restrictions?
> 
> ```
> [features]
> ...
> io_uring = ["rio"]
> 
> [dependencies]
> ...
> rio = { version = "0.9.2", optional = true }
> ```
> 
> Is there a guarantee that it will remain optional and that no required features will fall under a gpl license?

Oh, I definitely missed that too ^^'. I suppose that this means that, as long as you don't use this feature (or if you do, as long as you use the sponsors-only, MIT/Apache-2.0 version of `rio`), you can keep using `sled` like any other MIT/Apache-2.0 library.

PR #951, opened 2020-02-05
Make using the 'compression' feature easier

Byron:
That helps users who see the message to differentiate between a
generic compression feature, and that there is a feature called
compression.

It would have helped me anyway :).

Edit: And I figured out which compression levels may be used, so I documented that too.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -478,7 +478,7 @@ impl Config {
         if self.use_compression {
             supported!(
                 cfg!(feature = "compression"),
-                "the compression feature must be enabled"
+                "the 'compression' feature must be enabled"
             );
         }
         supported!(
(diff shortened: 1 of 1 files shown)


Byron:
I don't know why the build is failing though.

spacejam:
The windows one is a spurious one, but the linux crash failure seems fishy to me. I'm going to start burning that test on a server and investigating it further. In the mean time this code is safe to merge. Thank you for clarifying the docs for it!

Byron:
And thank you for creating such fantastic software! I am thoroughly impressed by the testing regimen employed here, which instills a whole lot of trust and is the reason for using a pre-release version of a database. Besides it has a wonderful and easy-to-use API with everything I need!

And it comes with some form of (optional) replication (which I believe is planned anyway), it will be perfect!
Something that usually prevents me from using distributed databases is that their configuration seems complicated, they seem to be unhappy in 'local' mode, and are slow to come up. Sled being 'embedded first' makes sure it will stay lean and not impose unnecessary burden.

Way to go!

spacejam:
Thanks for the kind words, @Byron :)

PR #952, opened 2020-02-06
segment accounting tune-up

spacejam:
this work aims to help address some of the space overuse issues by improving our understanding of space usage, as well as using better metrics to drive runtime storage garbage collection

- [x] new metrics around segment utilization
- [x] use size-based accounting
- [x] optimize segment GC with flat-combining and Arc-based ownership
- [x] debug space leak


spacejam:
measurements & naive implementation done in the naive and pessimistic way. initial checks show pretty nice space reductions, but we removed a fast path that bypasses a mutex in many cases. 

this blocking usage can be made totally non-blocking by leaning into flat-combining and tracking segment ownership in a lower contention way

Issue #397, opened 2018-11-02
OS-agnostic crash tests


spacejam:
We currently skip crash tests for Fuchsia, Android, and Windows because we rely on `libc::fork`, `libc::raise`, and `libc::waitpid` to create separate processes, crash them abruptly, and wait for the exit code from the parent. We may be able to replace all of these with functionality that is present in `std::process` in a platform-agnostic manner.

In the [current implementation](https://github.com/spacejam/sled/blob/f061b01856acee1ffa8381c8f4179d4c5090a30b/tests/tests/test_crash_recovery.rs), we could simply replace the call to `raise(9)` with `std::process::exit(9)`.

We can use `std::process::Command::status` to spawn a new process and wait on its exit code, and assert that it is `Some(9)`.

How we ensure a binary is available for the test to call is left up to the implementor!

roignpar:
Hello!
I would like to try and implement this.
I have never worked with C forking, but if I understand it correctly, [`std::process::Child`] works differently. As far as I can tell, the code found in [`run_without_snapshot`] and [`run_with_snapshot`] must be compiled and ran as separate binaries using [`Command::spawn`].

I currently have no clue how to build binaries from arbitrary code like [`run_without_snapshot`] or [`run_with_snapshot`] contain and run them from within a test.
Maybe move the functions to separate files, inside a `fn main()` and build + run those files using `Command`s, but I am not sure where these files should be placed, or even if this is a good approach.
Do you have any ideas or hints?

[`std::process::Child`]: https://doc.rust-lang.org/std/process/struct.Child.html
[`run_without_snapshot`]: https://github.com/spacejam/sled/blob/master/tests/test_crash_recovery.rs#L214
[`run_with_snapshot`]: https://github.com/spacejam/sled/blob/master/tests/test_crash_recovery.rs#L231
[`Command::spawn`]: https://doc.rust-lang.org/std/process/struct.Command.html#method.spawn

divergentdave:
Opting out of the standard test runner might help with this. Then, we could write a new main() that either handles all the test cases, including spawning and killing subprocesses, or, if it's a subprocess, runs the database with the appropriate configuration.

This change would help because we wouldn't need to build extra binaries for the test suite, the subprocesses wouldn't have test runner thread complicating things, and we could get out of relying on fork to start a subprocess midway through the program.

roignpar:
@divergentdave thanks for replying.
[...]

PR #953, opened 2020-02-07
Add OS agnostic crash tests

roignpar:
This is what I wrote so far for  #397.
I used the strategy described by @divergentdave.
Test names are also used as dir names to easily share them with child processes.
Haven't added the batches tests yet. Should I add them? They are currently ignored.
Renamed the old test file to have it around for reference. Will remove it once this is done.


divergentdave:
The batch crash tests are currently ignored because they are known to fail, (because batches can get corrupted with two consecutive crashes) and fixing them is an open issue. I'd recommend adding parallel tests for them, and ignoring them in the new integration test file.

roignpar:
@divergentdave I'm not sure what you mean by "parallel" tests.
Add them to the new test file, but don't run them in `main`?

divergentdave:
Sorry, didn't think that through. I guess you can't use the #[ignore] attribute in a file that doesn't use the test runner. At any rate, consider writing the code for doing crash tests with batches, and then comment out the lines that invoke it. Once batches have been redesigned later on, we can un-comment the lines to enable the test.

roignpar:
@divergentdave I added the batch tests (commented out).
I think these are the minimal changes needed in order to remove `libc` from the crash tests.
Is there anything else that should be changed?

spacejam:
Great work, @roignpar :] I'm excited to learn about how badly various operating systems start breaking due to poor fsync assumptions now xD 

I just fixed some merge conflicts caused by completely removing runtime snapshotting, which I did not trust very much.

roignpar:
@spacejam Thanks for fixing the merge conflicts.
I'm completely new to sled, but if you think I can help with other (easy-ish) issues, let me know!

PR #959, opened 2020-02-11
file GC-related refactoring

spacejam:
File GC has been pretty ad-hoc, bug-prone, and highly memory inefficient. This PR seeks to significantly reduce memory usage, CPU usage associated with deferred cleanups, and make the general tracking of segments more strongly typed by leaning into a stronger state machine approach.

- [x] switch segment tracking over to a more strictly specified state machine
- [x] switch over to using BTreeSet's in a few places to reduce space usage over the old VecSet/hash sets
- [x] debug
- [x] profile
- [ ] update SA comments to reflect current arch


spacejam:
this results in some nice memory reductions for a write-heavy workload of small updates that stressed out the segment accountant:

* master: 870mb allocated over workload
* this with hashsets: 771mb
* this with btreesets: 732mb (and this also ran fastest)

PR #961, opened 2020-02-11
significant contention reduction for file GC

spacejam:
a key aspect of reducing contention through the whole system is to reduce reliance on the SegmentAccountant mutex. writers need to check to see if they can help perform GC from time to time, and this lets them use a very low contention mutex connected to a mpsc::Receiver instead of the SegmentAccountant itself.

### DIFF
--- a/src/pagecache/iobuf.rs
+++ b/src/pagecache/iobuf.rs
@@ -1,8 +1,6 @@
 use std::{
-    cell::UnsafeCell,
-    sync::atomic::AtomicBool,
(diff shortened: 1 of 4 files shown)


spacejam:
This results in a ~10% throughput increase on workloads with high numbers of threads all trying to write at once.

* master workload w/ 12 threads: 1.7m writes/s, big accountant mutex section in flamegraph
* this: 1.9m writes/s, nice evenly spread flamegraph even with 12 threads

PR #972, opened 2020-02-20
use fixed-size Copy on-stack Vec for storing CacheInfo with fewer allocations

spacejam:


spacejam:
passed miri when run on extracted program

PR #982, opened 2020-02-22
Initial cargo-c support

lu-zero:
I added a small readme and the cbindgen configuration that seems working fine.


spacejam:
Thanks a lot for adding this, @lu-zero :) Excited that this is getting a nicer experience around it, since it's potentially so high impact for folks in ecosystems other than rust!

PR #985, opened 2020-02-23
Set up github actions to use more tests

spacejam:


spacejam:
will leave both azure & github actions running in parallel for a little bit to observe if they ever diverge. github actions is faster on the first parallel run. windows is the laggard in both by a bit, maybe it makes sense to reduce the tree quickcheck tests on windows.

PR #1010, opened 2020-03-06
std error for cas errors

### LINKED ISSUE #1009: Feature request: implement `std::error::Error` on `CompareAndSwapError`
I just noticed that [CompareAndSwapError](https://docs.rs/sled/0.31.0/sled/struct.CompareAndSwapError.html) doesn't implement [std::error::Error](https://doc.rust-lang.org/std/error/trait.Error.html).  Would you be willing to implement it?  It would make error propagation much easier via e.g. [anyhow](https://crates.io/crates/anyhow).

spacejam:
Closes #1009

### DIFF
--- a/src/tree.rs
+++ b/src/tree.rs
@@ -2205,9 +2205,4 @@ impl fmt::Display for CompareAndSwapError {
     }
 }
 
-impl std::error::Error for CompareAndSwapError {
-    #[allow(deprecated, deprecated_in_future)]
-    fn description(&self) -> &str {
-        "Compare and swap conflict"
-    }
-}
+impl std::error::Error for CompareAndSwapError {}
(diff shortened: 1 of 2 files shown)


src/tree.rs
```diff
@@ -2204,3 +2204,10 @@ impl fmt::Display for CompareAndSwapError {
         write!(f, "Compare and swap conflict")
     }
 }
+
+impl std::error::Error for CompareAndSwapError {
+    #[allow(deprecated, deprecated_in_future)]
+    fn description(&self) -> &str {
+        "Compare and swap conflict"
+    }
+}
```
Kerollmops: As I read [the description of the Error trait](https://doc.rust-lang.org/std/error/trait.Error.html#method.description), it is better to avoid using this method, and impl fmt::Dsiplay on the error type.

spacejam: Ahh I wasn't aware, in that case I'll close this. Thanks!

spacejam: Err, misunderstood, just removed `description`s and will just avoid that method.

Issue #1013, opened 2020-03-10
Change the api of transaction apply_batch


Erk-:
I propose to change the api of the `apply_batch` method in transactions from:
```rust
pub fn apply_batch(
    &self,
    batch: Batch,
) -> UnabortableTransactionResult<()>
```
to
```rust
pub fn apply_batch(
    &self,
    batch: &Batch,
) -> UnabortableTransactionResult<()>
```
This will then mean that it will be possible to write code like this:
```rust
let db = config.open().unwrap();

let t1 = db.open_tree(b"1")?;

let mut b1 = Batch::default();
b1.insert(b"k1", b"v1");
b1.insert(b"k2", b"v2");

t1.transaction(|tree| {
    tree.apply_batch(&b1)?;
    Ok(())
})?;
```

spacejam:
I like it! Thanks for improving this API :)

PR #1015, opened 2020-03-12
Change transaction `apply_batch` to take argument by-reference.

Erk-:
This change facilitates that you can setup a batch and then execute it
inside of a transaction. For example:

```rust
let db = config.open().unwrap();
let t1 = db.open_tree(b"1")?;

let mut b1 = Batch::default();
b1.insert(b"k1", b"v1");
b1.insert(b"k2", b"v2");

t1.transaction(|tree| {
    tree.apply_batch(&b1)?;
    Ok(())
})?;
```

----

This PR is the implementation of #1013 if it is accepted.

----

I did have a spurious fail in the test `log_writebatch`, but I am unsure if 
it is connected to this change in any way.

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -299,9 +299,9 @@ impl TransactionalTree {
     /// Atomically apply multiple inserts and removals.
     pub fn apply_batch(
         &self,
-        batch: Batch,
+        batch: &Batch,
     ) -> UnabortableTransactionResult<()> {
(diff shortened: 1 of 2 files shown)


spacejam:
@Erk- this looks good, thanks for improving this :) I'll add an entry to the CHANGELOG mentioning this breaking change for the upcoming `0.32` release.

Issue #818, opened 2019-09-10
move all wiki pages into doc module


spacejam:
similar to how https://docs.rs/ndarray/0.12.1/ndarray/doc/index.html has a documentation module, I think this is a nice place to keep the docs in a more properly version-controlled place.

Th3Whit3Wolf:
This seems pretty simple. I can help with this.

spacejam:
Thanks, @Th3Whit3Wolf :)

PR #1020, opened 2020-03-15
Doc modules

Th3Whit3Wolf:
This commit address [issue 818](https://github.com/spacejam/sled/issues/818) and moves all wiki pages into a doc module.

I did it the same way [ndarray](https://github.com/rust-ndarray/ndarray) does it with doc module building only with feature docs and setting doc.rs to build with that feature as well.


spacejam:
Nice work, @Th3Whit3Wolf :) Thank you for this!

Th3Whit3Wolf:
No problem! Thanks for the awesome project!

PR #1021, opened 2020-03-20
cleanups, more clippy lints, replace fail crate with stale deps with simple internal impl

spacejam:


spacejam:
failing repeatedly on ICE with docopt, and docopt caused a bunch of latency during compiles due to pulling in serde, so I just replaced docopt with a very simple arg parser that doesn't change functionality at all

PR #1025, opened 2020-03-27
implement stateful merge operators

### LINKED ISSUE #1022: Feature: stateful merge operators
It's sometimes useful to have merge operators that are able to access stuff other than the data relating to the currently modified key, be it app configuration, serializers, loggers, etc. With the current API, this has to be done by accessing the global state, which is a rather poor solution. I experimented a little bit in my fork, and it seems to be a pretty minimal change. However it would require changing the API slightly:

- `MergeOperator` would have to become a `trait` / trait alias for `Fn(&[u8], Option<&[u8]>, &[u8]) -> Option<Vec<u8>>`.
- `Tree` would have to store a boxed `MergeOperator` trait object. `set_merge_operator` isn't usually called often, so the allocation there shouldn't matter. Still, something like [`smallbox`](https://crates.io/crates/smallbox) could be used to not heap-allocate small closures. 
- I couldn't find the merge functionality in the C API, but that could be relatively easily done using the usual method of passing data pointer.

mrobakowski:
Closes #1022

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -365,8 +365,5 @@ pub(crate) type FastSet8<V> = std::collections::HashSet<
 
 /// Allows arbitrary logic to be injected into mere operations of the
 /// `PageCache`.
-pub type MergeOperator = fn(
-    key: &[u8],
-    last_value: Option<&[u8]>,
-    new_merge: &[u8],
-) -> Option<Vec<u8>>;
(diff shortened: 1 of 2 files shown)


spacejam:
LGTM :) Thank you for this update! I'll include it in the next release

PR #1027, opened 2020-04-04
FFI Improvements

bltavares:
Hello there,

I've spent some hours today trying out how to put [sled on Flutter](https://github.com/bltavares/treno/pull/1) using FFI.

I was able to get `sled-native` to create the DB, but I'm not yet being able to call `get` nor `set`, but it seems to be mostly on the side of Dart.

This commits are some set of changes I had to apply to make it more ergonomic on Dart, and possible to call the `sled-native` library.

One special detail is that `sled_confg_set_path` was not working for me, as it was using the internal Arc counter to control whether or not it was in use and calling `.as_mut()` panicked as it was already cloned.

This is the biggest change on the code. I'm not sure if it is sound tho, as I know very little about C or FFI. I hope they are safe :)


bltavares:
I also found it would not compile with `cross build --target aarch64-linux-android` due to some changes on how `CStr` is defined, on `as *mut i8` expecting it to be `as *mut u8`


```
   = note: expected raw pointer `*mut u8`
              found raw pointer `*mut i8`
```

I'm not sure how to fix this tho.

bltavares:
With this last commit, I was able to `get` and `set` from Flutter. I didn't test the iter and scan methods yet.

<details>
<summary> Screenshot</summary>

Screen Shot 2020-04-04 at 1 30 12 AM

</details>

spacejam:
@bltavares thanks for this :) I've changed a few things, to avoid returning the same pointer that is passed in when working with a Config object. Could you confirm if this still works for you?

bltavares:
@spacejam I've tried this changes before, but it does not compile.

The `Config` calls moves the ownership of the config and returns the struct again, that is why I ended up doing the roundtrip.

```
error[E0507]: cannot move out of `*config` which is behind a raw pointer
  --> src/lib.rs:49:5
   |
49 |     (*config).path(value);
   |     ^^^^^^^^^ move occurs because `*config` has type `sled::config::Config`, which does not implement the `Copy` trait

error[E0507]: cannot move out of `*config` which is behind a raw pointer
  --> src/lib.rs:58:5
   |
58 |     (*config).read_only(read_only == 1);
   |     ^^^^^^^^^ move occurs because `*config` has type `sled::config::Config`, which does not implement the `Copy` trait

error[E0507]: cannot move out of `*config` which is behind a raw pointer
  --> src/lib.rs:67:5
   |
67 |     (*config).cache_capacity(capacity as u64);
   |     ^^^^^^^^^ move occurs because `*config` has type `sled::config::Config`, which does not implement the `Copy` trait

error[E0507]: cannot move out of `*config` which is behind a raw pointer
  --> src/lib.rs:76:5
   |
76 |     (*config).use_compression(use_compression == 1);
   |     ^^^^^^^^^ move occurs because `*config` has type `sled::config::Config`, which does not implement the `Copy` trait

error[E0507]: cannot move out of `*config` which is behind a raw pointer
  --> src/lib.rs:86:5
   |
86 |     (*config).flush_every_ms(val);
   |     ^^^^^^^^^ move occurs because `*config` has type `sled::config::Config`, which does not implement the `Copy` trait
```

Also, the last change includes `IVec`, but does not import it, making it also not compile:

```diff
-use sled::{Config, Db, Iter};
+use sled::{Config, Db, IVec, Iter};
```

bltavares:
Could there be changes on `Config` API to take a `*mut Config` instead of `Config`? I'm not sure what's the impact on that API change tho.

spacejam:
ahh yes, the Config methods should take &mut self instead of self, I'll make this change

spacejam:
sorry for the confusion, Config is broken

bltavares:
No problems :)

I can wait for the changes on Config and fix this PR later

Em seg, 13 de abr de 2020 08:19, Tyler Neely <notifications@github.com>
escreveu:

spacejam:
I realized I was wrong about Config, because by having it consume and return owned versions, it allows users to rely on the builder pattern a bit more flexibly given other usage of it in some places. Sorry again for this taking so long to properly look at an merge. Thank you for cleaning this up :)

PR #1030, opened 2020-04-09
Various code cleanup things

### LINKED ISSUE #999: Minor API proposal: change `From` to `Into`
Several `Tree` methods (insert being the most obvious) have an interface like this:

```rust
fn modify<V>(&self, value: V)
    where IVec: From<V> { ... }
```

I propose changing these to `where V: Into<IVec>`, for the following reasons:

- It allows a *slightly* larger set of input types, as there are rare cases where `Into<T>` can be implemented but `T: From<U>` cannot.
- It allows the trait bound to be on `V`, the input parameter type, which personally I find to be cleaner than an indirect bound on `IVec`.
- Generally, using `Into<T>` is more common in Rust APIs [citation needed].
- This is a strictly forwards-compatible change, because `T: From<U>` implies `U: Into<T>`.

I'm happy to go through and make this change myself, but I wanted to first propose the change and get feedback / approval.

Lucretiel:
- Converted all instances of `T: From<U>` to `U: From<T>` (Fixes #999)
- rustfmt
- converted some instances of `if let` to a more idiomatic match
- Added Eq/PartialEq to ShutdownState

My editor automatically runs rustfmt but I can revert those changes if you prefer

### DIFF
--- a/examples/structured.rs
+++ b/examples/structured.rs
@@ -190,7 +190,8 @@ fn hash_join(db: &sled::Db) -> sled::Result<()> {
     let mut join = std::collections::HashMap::new();
 
     for name_value_res in &cats {
-        // cats are stored as name -> favorite_number + battles_won + home name variable bytes
+        // cats are stored as name -> favorite_number + battles_won + home name
+        // variable bytes
         let (name, value_bytes) = name_value_res?;
         let (_, home_name): (LayoutVerified<&[u8], CatValue>, &[u8]) =
             LayoutVerified::new_from_prefix(&*value_bytes).unwrap();
@@ -200,7 +201,8 @@ fn hash_join(db: &sled::Db) -> sled::Result<()> {
     }
 
     for name_value_res in &dogs {
-        // dogs are stored as name -> home name variable bytes + woof count + postal code
+        // dogs are stored as name -> home name variable bytes + woof count +
+        // postal code
         let (name, value_bytes) = name_value_res?;
 
         // note that this is reversed from the cat example above, where
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -35,18 +35,12 @@ pub struct Batch {
 
 impl Batch {
     /// Set a key to a new value
-    pub fn insert<K, V>(&mut self, key: K, value: V)
-    where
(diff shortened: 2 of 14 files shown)


Lucretiel:
It's not clear to me what's going on with the clippy failure; seems like something to do with GitHub actions?

spacejam:
Thanks for adding this! 

One small nit, could you switch the match statements on options back to `if let` blocks? I prefer to use `if let` anywhere that a single branch requires destructuring, because it reduces rightward pressure. I prefer to only use matches when there are multiple patterns that are interesting.

spacejam:
And yeah it seems like GH actions is having a lot of trouble :/

Lucretiel:
> I prefer to use if let anywhere that a single branch requires destructuring, because it reduces rightward pressure. I prefer to only use matches when there are multiple patterns that are interesting.

I'm not sure what you mean; all the cases that I changed have both a `Some` branch and a `None` branch. I didn't touch the branches that were `if let Some(x) == value { abc };`

spacejam:
@Lucretiel None requires no destructuring so it just adds an extra tab of rightward pressure, whereas the if-let allows full destructuring of all relevant information without the extra tabs

spacejam:
```
if let Some(interesting) = option {
    // one tab
} else {
    // one tab
}
```

vs

```
match option {
    Some(interesting) => {
        // two tabs
    },
    None => {
        // two tabs
    }
}
```

match adds a lot of extra line noise in this case.

does that make sense?

Lucretiel:
Yeah, totally, though in practice there often isn't actually a second tab, because it doesn't require braces at all:

```rust
if let Some(thing) = opt {
    Node::from_thing(thing)
} else {
    Node::from_none()
}
```
vs
```rust
match opt => {
    Some(thing) => Node::from_thing(thing),
    None => Node::from_none(),
}

I'll fix the ones that create rightward indentation pressure?

spacejam:
Since at this point conflicts are popping up, would you mind opening a small PR that only introduces the discussed Into/From changes? It adds a lot of friction to open a PR with several separate issues that are not related.

PR #1040, opened 2020-04-21
Fully initialize the end of IO buffers

### LINKED ISSUE #1032: Segment padding uninitialized memory problems
This is a follow-up from https://github.com/spacejam/sled/pull/937#issuecomment-612686235. To recapitulate, `IoBufs::write_to_log()` does no padding if there is less than 32 bytes available at the end of a segment, and then uses the full length of the segment when writing, which can result in a read from uninitialized memory.

In trying to write a patch for this, I ran into further conundrums. While I haven't tried to reproduce this yet, it seems that `IoBufs::write_to_log()` might not pad the entirety of the segment when it does pad. First, we have:

```rust
let pad_len = capacity - bytes_to_write - MAX_MSG_HEADER_LEN;
...
let padding_bytes = vec![MessageKind::Corrupted.into(); pad_len];
```

The `Cap` message header is written from offset `bytes_to_write` to `bytes_to_write + header_bytes.len()`. Then, the zeros are written from offset `bytes_to_write + header_bytes.len()` to `bytes_to_write + header_bytes.len() + pad_len`. Cancelling terms, the range of bytes that is written ends at offset `capacity + header_bytes.len() - MAX_MSG_HEADER_LEN`. We know that `header_bytes.len()` will be less than `MAX_MSG_HEADER_LEN`, because the `len` field will have a zero in its top byte, thus there will be uninitialized bytes at the end of the buffer. Moving through the function, if `should_pad` is true, then `maxed` was true, and thus `total_len` is set to `capacity`, and the uninitialized bytes will be read from and written to disk.

Correcting this issue looks like it would require solving a Catch-22, because making the `Cap` message extend to the end of the segment would require knowing the length of the `Cap` message's header, but that would in turn require knowing how long we wanted the message to be, etc. Perhaps this could be solved by using `capacity + header_bytes.len() - MAX_MSG_HEADER_LEN` as the message's length in its header, checksumming that many bytes in the message, but then zeroing out more bytes until the end of the segment.

A parallel concern is whether the read paths properly handle the red zone of each segment. I noticed that `LogIter` checks if `cur_lsn` is within `MAX_MSG_HEADER_LEN` of the end of a segment, and moves to the next segment. It appears that `IoBufs::start()`, one of the few other callers of `read_message()`, does not check for this case. Is such logic required here?

At this point, I'm feeling a little lost in the weeds, would appreciate input on how to tighten up segment writing/reading here.

divergentdave:
This writes bytes past the end of the last message in a segment (`Cap` or otherwise) to fully initialize the buffer. Initializing these bytes avoids UB from passing a partially initialized `&[u8]` to `pwrite_all()`.

Fixes #1032.

### DIFF
--- a/src/pagecache/iobuf.rs
+++ b/src/pagecache/iobuf.rs
@@ -554,12 +554,8 @@ impl IoBufs {
         // a pad is a null message written to the end of a buffer
         // to signify that nothing else will be written into it
         if should_pad {
-            let pad_len = capacity - bytes_to_write - MAX_MSG_HEADER_LEN;
-            let data = iobuf.get_mut_range(bytes_to_write, pad_len);
-
-            // take the crc of the random bytes already after where we
-            // would place our header.
-            let padding_bytes = vec![MessageKind::Corrupted.into(); pad_len];
+            let pad_len = unused_space - MAX_MSG_HEADER_LEN;
+            let data = iobuf.get_mut_range(bytes_to_write, unused_space);
 
             let segment_number = SegmentNumber(
                 u64::try_from(base_lsn).unwrap()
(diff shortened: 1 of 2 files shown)


spacejam:
Great work, @divergentdave :) This was a pretty legendary bugfix, considering all of the effort that has gone into miri support to discover the bug! I'm bragging about your work on this when I'm talking with friends, by the way! There's a blog post that will feature some of this that is in the works, and I'll keep you in the loop on that so you can get final sign-off on anything that talks about your work.

PR #1041, opened 2020-04-23
safety-critical recovery approach

spacejam:
- [x] merging @divergentdave's batch failpoint work
- [x] accounting for batches in the log stabilization code
- [ ] more formally document emergent properties of the system which must be controlled for in a documented way, and tested in a documented way
  - fully atomic serializable batch recovery
- [ ] more formally document requirements around file reuse, stability and crash recovery
  - a segment may only be reused after it is no longer in the unstable tail
  - a segment may only be reused after the segments that replaced its data are no longer in the unstable tail
  - strictly monotonic segments. never duplicate segments
- [x] ~~make failpoints work with a "bit queue" so that they may not be triggered upon first hit of the failpoint~~ #1052 
- [ ] separate synchronous requirements from asynchronous requirements, allow async requirements to be disabled to ensure they don't cause bugs to happen by being async
  - runtime segment reuse
  - recovery segment freeing
  - snapshot reading (only can be disabled in combination with segment freeing/reuse)
- [x] uncomment page_out body and clear stability assumptions

### DIFF
--- a/src/pagecache/iobuf.rs
+++ b/src/pagecache/iobuf.rs
@@ -829,16 +829,21 @@ impl IoBufs {
         if let Some(new_stable_lsn) = updated {
             self.stable_lsn.store(new_stable_lsn, SeqCst);
 
+            #[cfg(feature = "event_log")]
+            {
+                // We add 1 because we want it to stay monotonic with recovery
+                // LSN, which deals with the next LSN after the last stable one.
+                // We need to do this while intervals is held otherwise it
+                // may race with another thread that stabilizes something
+                // lower.
+                self.config.event_log.stabilized_lsn(new_stable_lsn + 1);
+            }
+
             // having held the mutex makes this linearized
             // with the notify below.
             drop(intervals);
 
             let _notified = self.interval_updated.notify_all();
-
-            #[cfg(feature = "event_log")]
-            {
-                self.config.event_log.stabilized_lsn(new_stable_lsn);
-            }
         }
     }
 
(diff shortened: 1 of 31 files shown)


spacejam:
merging this large number of bugfixes and refactors now that the known bugs have been addressed, will follow up later with refined safety requirements and STPA-driven analysis based on the many bugs found through this effort

src/pagecache/iobuf.rs
```diff
@@ -191,10 +191,104 @@ impl IoBuf {
...
+            } else {
+                break;
+            }
+        }
+
+        if self.batches.is_empty() {
+            Some(self.stable_lsn)
         } else {
-            Err(res)
+            batch_stable_lsn
```
divergentdave, line 291: It looks like this method will return None if there is one batch being tracked and it is not yet stable. Is that intended, or would it make sense to return the lesser of self.stable_lsn or one less than the first batch's low LSN?

divergentdave: I suppose the more general version of that question would be, why is batch_stable_lsn set based on the end of a stabilized batch, rather than just before the beginning of the first non-stable batch? (I may be making some faulty assumptions about what the variables mean)

spacejam: that's a good point, I think it should be bumped up to just before the earliest batch if that's less than or equal to the internal contiguous stable point. I should also improve the comments in that area. 

A goal is to put this into a more formal model of constraints that can be linked down to code, where things like this can be reasoned about more mechanically so stuff like this will hopefully shake out in these correctness critical pathways

Issue #1045, opened 2020-04-26
Use of uninitialized memory in Debug impl for StackVec


divergentdave:
1. sled version: master branch
2. rustc version: rustc 1.44.0-nightly (9b2f8dbba 2020-04-20)
3. operating system: Linux

#1042 added an implementation of `Debug` for `StackVec`. The function transmutes an array of `MaybeUninit<CacheInfo>` to `CacheInfo`, ignores the length field, and passes the array as a `&dyn Debug` trait object to be formatted as well. I haven't chased down if there are any UB-causing uses of this implementation in the code base, but the following one-liner demonstrates use of uninitialized memory:

```rust
dbg!(StackVec::default());
```

The standard library's `impl<T> Debug for MaybeUninit<T>` just prints its own type name. (https://doc.rust-lang.org/src/core/mem/maybe_uninit.rs.html#238-242) `impl Debug for StackVec` should be changed either to do that, or to piggy-back off of `deref()` and only pass the initialized slice of the backing array.

spacejam:
@divergentdave great find, thank you for fixing this :)

PR #1046, opened 2020-04-26
Rewrite Debug implementation for StackVec

divergentdave:
Fixes uninitialized memory use by only printing the first `len` entries of the backing array (by using `deref()`). Fix for #1045. [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=809f6670dae846ce33a659873d2fb399), for kicking the tires on the formatted output.


spacejam:
Thanks!

Issue #1048, opened 2020-04-29
set failpoints that trigger after multiple hits


spacejam:
currently failpoints will trigger the first time they get hit after the failpoint has been activated. this hides bugs that only get triggered on crashes during subsequent calls. we can change failpoints to being a simple u64 with some number of bits set. each time the failpoint gets hit, we can bitshift and pop the bit off, effectively treating it as a queue of bools.

PR #1052, opened 2020-05-03
Trigger failpoints randomly on calls after set

divergentdave:
This implements #1048, associating a randomly generated u64 with each failpoint, and checking a different bit each time the failpoint is passed when deciding whether to return an error. I also included special handling for shrinking test cases, by progressively clearing bits in the bitset.

I threw several core-hours at this overnight (using `QUICKCHECK_TESTS=1000000 QUICKCHECK_MAX_TESTS=1000000`) and I got no crashes, and one deadlock with "log flusher failed to make_stable after 30 seconds." I have my suspicions that the deadlock has been a latent problem, predating this PR.

### DIFF
--- a/src/fail.rs
+++ b/src/fail.rs
@@ -1,26 +1,36 @@
-use std::collections::HashSet;
+use std::collections::HashMap;
 
 use parking_lot::Mutex;
 
 use crate::Lazy;
 
-type HS =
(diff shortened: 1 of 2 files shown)


spacejam:
Nice work! Thank you for implementing this :)

PR #1071, opened 2020-05-19
Fix histogram test

divergentdave:
I noticed the histogram tests have bitrotted. This fixes one test to reflect the change in return type, and adds `--no-default-features` to a command line so that the histogram tests can run in the CI.


spacejam:
nice catch! thanks :)

Issue #1070, opened 2020-05-19
MIPS support


bltavares:
Hi there,

I've recently tried to [compile a project](https://github.com/Kuska-ssb/solar/issues/6) that depends on `sled` to [OpenWRT](https://openwrt.org/) mips device using [cross](https://github.com/rust-embedded/cross).

```
 cross build --target mips-unknown-linux-musl
```

This led me to type errors similar to the ones found on [async-std](https://github.com/async-rs/async-std/pull/756) for 32bit devices.

I've tried to locally change the types from `AtomicU64` to `AtomicUsize` as they did there, and there were many more errors to fix...

I'm curious if there is interest on having the project working on MIPS devices, and it I could give it a try to get it compiling. Cheers.

bltavares:
[This is certainly wrong](https://github.com/bltavares/sled/commit/c40df34217e42f3ed0eadaa017a390d50544bcf1) but it still compiles on x64, but it made some good progress compiling for mips.

I would love to hear If you have any recommendations on the way to go, or even a guideline on how to cast things. I've tried following the compiler errors as casting to make it fix hehe

divergentdave:
I think the underlying issue isn't the processor word size, but the absence of the atomic memory instructions. (sled works on i686 as-is) The docs say AtomicU64 isn't supported on 32-bit MIPS or PowerPC. https://doc.rust-lang.org/std/sync/atomic/index.html#portability AtomicU64 is baked in pretty deeply to the architecture of sled, you may notice that the IoBuf header actually holds multiple bit-packed fields, for example.

bltavares:
Thanks for replying so fast! As you can notice, I don't know much about MIPS or Atomic Instructions.

Given that only two tests failed on the serialization part, do you think it is even worth to try to make it pass, given you mentioned that AtomicU64 is core to how sled works?

I know MIPS is quite niche, and I have only a passer-by interest at this moment, with at most a hobbyist application in the far future (eg: embed a db for wifi configs). I would be ok to close the issue if it is too far fetch to support it.

spacejam:
does Mutex exist for your MIPS target? If so, we may have nice luck by lust conditionally compiling an AtomicU64 for MIPS that is the underlying 64-bit item wrapped in a Mutex, and just ignores the Ordering argument that is passed to its methods. It may also be nice for doing things like asserting that (certain, all?) CAS operations never fail when testing sled in single-threaded mode.

bltavares:
[...]

PR #1076, opened 2020-05-20
Port AtomicU64 to compile for MIPS

bltavares:
Rust's `std::sync` namespace does not provide `AtomicU64` and
`AtomicI64` for MIPS and PowerPC architectures.

This make the project not capable of being embedded into routers as
mentioned on #1070

This commit changes the references of `AtomicU64` and `AtomicI64` to use
a portable version, wich fallbacks to `crossbeam` Mutex when Atomics are
not available.

To compile against MIPS, the `block` initialization had to change, as
the Mutex is not `Copy`, and Rust's initialization of arrays only
supports `Copy` types. The `array-init` crate supports non-Copy types,
and that is why it was introduced.

It also introduces a `--features mutex` crate, which allows to compile
to x64 or any other architecture using the Mutex implementation, useful
for testing without cross-compiling.

Tests passes with:

```
cargo test --features testing
cargo test --features testing --features mutex

cross build --target mips-unknown-linux-musl
```

Cross-compiling tests fails due to sub-process spawning not getting
hooked on qemu binfmt setup and spawning with the wrong architecture
(x64 container trying to run MIPS executable).

Creating a MIPS Debian VM using QEMU compiles (although very slowly) and
runs almost all of the tests correctly (except the quiescent cpu time, which
is very slow to run on my VM).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ measure_allocs = []
 pretty_backtrace = ["color-backtrace"]
 io_uring = ["rio"]
 docs = []
+mutex = []
 
 [dependencies]
 crossbeam-epoch = "0.8.2"
@@ -47,6 +48,7 @@ parking_lot = "0.10.2"
 color-backtrace = { version = "0.4.2", optional = true }
 rio = { version = "0.9.3", optional = true }
 backtrace = "0.3.49"
+array-init = "0.1.1"
 
 [target.'cfg(any(target_os = "linux", target_os = "macos", target_os="windows"))'.dependencies]
 fs2 = "0.4.3"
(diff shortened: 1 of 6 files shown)


spacejam:
Sorry for the very slow feedback. I generally like the idea of this a lot, but I would prefer not to rely on a new external dependency for this functionality. Would you mind making the optional shim a module in the src directory?

spacejam:
LGTM! Thank you very much for adding this :) 

Again, sorry for the very slow feedback, I was out in the mountains for a bit, but am back now

bltavares:
Awesome. I'm glad to help and I hope you had a good time out in the mountains :)

PR #1078, opened 2020-05-24
Default method of making multi-tree transaction is very rigid (uses t…

pantonov:
Default method of making multi-tree transaction is very rigid (uses tuples of trees), so it wasn't possible to build multi-tree transactions incrementally. Also, public trait `Transactional<E>` weren't usable at all without ability for crate users to construct `TransactionalTree` and `TransactionalTrees`.

1. Added pub constructor `TransactionalTree::from_tree(&Tree) -> Self` (this also removes tiny bit of code duplication)
2. Added pub constructor `TransactionalTrees::from_vec(Vec<TransactionalTree>)`
3. Implemented `From<UnabortableTransactionError> for TransactionError` for usability reasons. Conversion of `UnabortableTransactionError::Conflict` to `TransactionError` conversion panics, returning `Conflict`  outside of transaction is a logic error.
4. `Transactional::view_overlay` now takes `&self`, this allows passing arbitrary stuff to transaction closure with custom implementations of `Transactional` trait.

Example provided in documentation of  `TransactionalTrees`, as below:

```rust
use sled::{ transaction::*, Config };

fn main() -> TransactionResult<()> {
    let config = Config::new().temporary(true);
    let db1 = config.open().unwrap();
    let dba = db1.open_tree(b"a").unwrap();
    let dbb = db1.open_tree(b"b").unwrap();

    struct MyTransactions {
        trees: Vec<TransactionalTree>,
        my_stuff: i32,
    }
    impl Transactional<()> for MyTransactions {
        type View = i32;
        fn make_overlay(&self) -> TransactionalTrees {
            TransactionalTrees::from_vec(self.trees.clone())
        }
        fn view_overlay(&self, _: &TransactionalTrees) -> Self::View {
            self.my_stuff
        }
    }

    let mut my_transactions = MyTransactions{ trees: vec![], my_stuff: 111 };

    // operations on TransactionalTree cannot cause conflicts until fed to transaction()
    let tt1 = TransactionalTree::from_tree(&dba);
    tt1.insert(b"k1", b"cats")?;
    tt1.insert(b"k2", b"dogs")?;
    my_transactions.trees.push(tt1);


    let tt2 = TransactionalTree::from_tree(&dbb);
    tt2.insert(b"k3", b"bats")?;
    tt2.insert(b"k4", b"rats")?;
    my_transactions.trees.push(tt2);

    let ret = my_transactions.transaction(|my_stuff| Ok(my_stuff + 1))?;
    assert_eq!(ret, 112);

    assert_eq!(&dba.get(b"k1")?.unwrap(), b"cats");
    assert_eq!(&dba.get(b"k2")?.unwrap(), b"dogs");
    assert_eq!(&dbb.get(b"k3")?.unwrap(), b"bats");
    assert_eq!(&dbb.get(b"k4")?.unwrap(), b"rats");

    Ok(())
}
```

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -141,6 +141,17 @@ impl<E> From<UnabortableTransactionError> for ConflictableTransactionError<E> {
     }
 }
 
+impl<E> From<UnabortableTransactionError> for TransactionError<E> {
+    fn from(error: UnabortableTransactionError) -> Self {
+        match error {
+            UnabortableTransactionError::Conflict =>
+                panic!("UnabortableTransactionError::Conflict called outside of transaction"),
+            UnabortableTransactionError::Storage(other) =>
+                TransactionError::Storage(other)
+        }
+    }
+}
+
 /// An error type that is returned from the closure
 /// passed to the `transaction` method.
 #[derive(Debug, Clone, PartialEq)]
@@ -242,6 +253,14 @@ impl<T> From<Error> for TransactionError<T> {
 }
 
 impl TransactionalTree {
+    /// Create new `TransactionalTree` wrapper from `Tree`
+    pub fn from_tree(tree: &Tree) -> Self {
+        Self {
+            tree: tree.clone(),
+            writes: Default::default(),
+            read_cache: Default::default(),
+        }
+    }
     /// Set a key to a new value
     pub fn insert<K, V>(
         &self,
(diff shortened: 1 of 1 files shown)


spacejam:
It's intentional for TransactionalTrees to be privately constructed because by making this public, it starts to bolt the implementation to a lot of assumptions with the concurrency control path.

The reason why this originally operated over tuples instead of lists was because it seemed likely that trees would become typed soon after. This is less likely now, although different kinds of type-friendliness will be added.

I would be open to a PR that implements Transactional for `[Tree]` though, where the slice of trees gets mapped to the transactional component.

I intentionally do not want to expose users to any constructors for TransactionalTrees/TransactionalTree because it starts to get into important concurrency control considerations that they should never have direct control over.

spacejam:
I appreciate the work you put into this, but please get in touch earlier in the process, because I know it sucks to have hard work rejected, and there are definitely places in the implementation that I see as useful improvements.

pantonov:
Hi,

I think basically Transactional for [Tree] should solve my particular issue
(dealing with zillon tuple type signatures), I'll play with it and if it
goes well will get back with a new PR. I'll also add private constructor
for TransactionalTree, just to reduce code repetition in Transactional
impls.

-- Paul

PR #1079, opened 2020-05-24
Implemented Transactional for [Tree] and [&Tree]

pantonov:


pantonov:
An example use-case: imagine we have some objects A and B, each wraps several Trees of the same Db. We need to write some generic method or macro C, which abstracts over A, B... and binds updates to A and B into the single transaction. With current state of tuple types support in Rust (no generic iterators over tuples, etc) that's difficult to implement without using slices.

spacejam:
This looks good, thank you for implementing this! I think the 0.32 release will be out this weekend if the last debugging effort goes well, and this change will be included in it. 

I believe the test failure is something I've addressed already in #1041, and I'll handle any clippy issues and the CHANGELOG update post-merge.

PR #1083, opened 2020-05-30
Drop mutex earlier to fix no-threads configuration

divergentdave:
When running tests with configuration that disables the threadpool, I was getting a deadlock on the intervals mutex from anything that used transactions. The transaction commit eventually calls `Reservation::mark_writebatch()`, which locks the intervals mutex, calls `StabilityIntervals::mark_batch()`, and calls `Reservation::complete()`. The issue is that `Reservation::complete()` eventually spawns a threadpool task that calls `IoBufs::write_to_log()`, which also needs to lock the intervals mutex. If the threadpool is disabled, this closure will be called right away, and `write_to_log` will deadlock on the mutex.

This PR resolves the issue by dropping the intervals mutex guard before calling `Reservation::complete()`. This is a fairly naive fix, and I'm not familiar with recent related changes, is it okay to unlock the mutex here?

### DIFF
--- a/src/pagecache/reservation.rs
+++ b/src/pagecache/reservation.rs
@@ -103,6 +103,7 @@ impl<'a> Reservation<'a> {
 
             let mut intervals = self.log.iobufs.intervals.lock();
             intervals.mark_batch((self.lsn, peg_lsn));
+            drop(intervals);
 
             self.complete()
         }


spacejam:
Great catch, nice work!

spacejam:
Yeah, my understanding is that this mutex is not necessary for upholding any invariants around batches. batches will be applied while holding an exclusive writer lock mutex, but this is upheld from the concurrency control module and it's required to be held for the duration of this call. I'm about to merge Guard with the concurrency control object to enforce a valid concurrency control object is being held during things like this.

spacejam:
Hmm but there's a chance it was unintentionally load-bearing haha... https://github.com/spacejam/sled/pull/1084/checks?check_run_id=723391335

spacejam:
will dive in

PR #1086, opened 2020-05-31
Allow stabilization to happen up to the beginning of the lowest write…

spacejam:
…batch. Add an additional assertion to the stabilization code.


spacejam:
cc @divergentdave

PR #1088, opened 2020-05-31
[978] Fuzzy checkpointing

blitzerr:
Convert the PageTable into a vector of PageState.

### DIFF
--- a/src/pagecache/mod.rs
+++ b/src/pagecache/mod.rs
@@ -432,7 +432,6 @@ pub struct Page {
 }
 
 impl Page {
-    #[allow(unused)]
     pub(crate) fn to_page_state(&self) -> PageState {
         let is_free =
             if let Some(Update::Free) = self.update { true } else { false };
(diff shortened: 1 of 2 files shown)


src/pagecache/mod.rs
```diff
@@ -2076,4 +2075,37 @@ impl PageCache {
...
+        let pid_bound_usize = assert_usize(pid_bound);
+
+        let mut page_states =
+            Vec::<PageState>::with_capacity(pid_bound_usize);
+        let guard = pin();
+        for pid in 0..pid_bound {
+            if let Some(pg_view) = self.inner.get(pid, &guard) {
+                let page_state = pg_view.to_page_state();
+                page_states.push(page_state);
+            }
```
spacejam, line 2102: this looks great overall! there's just two subtle race conditions I realized though while reading this:

1. some other thread is going to allocate a new page, so they bump the `next_pid_to_allocate`. they have not yet installed anything into the `PageTable` though, so if anyone were to try to get the page, it will return `None`.
2. the allocating thread installs an empty pagetable entry before calling cas_page on it to actually install the data. if this thread calls to_page_state on it, it will panic due to array indexing 0 on an empty array. 

I'll fix both of these by looping until both of the above conditions are false, since the allocating thread will fix both of them on its own and we don't have the ability to cooperatively complete their operation without knowing the page they are trying to allocate.

These don't need to be fixed in this PR, I'll do it after merging.

spacejam:
the failing test has to do with a crash bug that I need to address and then cut a 0.32.0-rc2

yojs:
The races are fixed by @spacejam [here](https://github.com/spacejam/sled/pull/1089).

src/pagecache/mod.rs
yojs: Sounds good

PR #1092, opened 2020-06-03
refactor shutdown logic. clear some deprecated items. properly spin-up the flusher thread on the BSDs

spacejam:
@Erk- this may have an effect on some of the size issues you were seeing on FreeBSD, since I believe we may not have actually been spinning up the thread (still). Damn I hate conditional compilation.


Erk-:
I will give it a attempt later today and see if it runs better.

PR #1095, opened 2020-06-05
runtime memory improvements

spacejam:
remove StackVec
use Option<NonZeroU64> in Node pointers
Box the updates for paged-in items


spacejam:
before: 

image

after:
massif_sled_after

PR #1096, opened 2020-06-06
Custom Arc without weak count to reduce memory usage

spacejam:


src/arc.rs
```diff
@@ -0,0 +1,215 @@
...
+        let ptr = Box::into_raw(bx);
+        Arc { ptr }
+    }
+
+    pub fn copy_from_slice(s: &[T]) -> Arc<[T]> {
+        let align =
+            std::cmp::max(mem::align_of::<T>(), mem::align_of::<AtomicUsize>());
+
+        let rc_width = std::cmp::max(align, mem::size_of::<AtomicUsize>());
+        let data_width = mem::size_of::<T>().checked_add(s.len()).unwrap();
```
divergentdave, line 45: I think this should be a checked_mul here

src/arc.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+impl<T: Default> Default for Arc<T> {
+    fn default() -> Arc<T> {
+        Arc::new(T::default())
+    }
+}
+
+impl<T: ?Sized> Clone for Arc<T> {
+    fn clone(&self) -> Arc<T> {
```
divergentdave, line 129: Cloning should increase the reference count. (currently rc only gets initialized, read, and decremented)

spacejam: not sure how that slipped my mind during the first pass hahaha

src/arc.rs
spacejam: good catch, I think it's currently only used for u8's which is why that part didn't blow up yet

spacejam:
This results in roughly 5% fewer bytes allocated on a workload that writes 1 million items to the database (10b keys which get inlined in IVec, 100b values which use this Arc). For both workloads, there tend to be 97 bytes per allocation on average, but with this code there are 5% fewer allocations for some reason. 

So, this seems to be an improvement, but perhaps not for the reason expected.

divergentdave:
I'm getting an out-of-bounds memory read when running this new code through Miri. Here's a minimal test case I hacked together. (had to make stuff public and ignore lints)

```
use sled::arc::Arc;
  
fn main() {
    let a: Arc<[u8]> = Arc::copy_from_slice(b"1");
    println!("{}", a.len());
}
```

```
   --> /home/david/Code/sled/src/arc.rs:215:18
    |
215 |         unsafe { &(*self.ptr).inner }
    |                  ^^^^^^^^^^^^^^^^^^ memory access failed: pointer must be in-bounds at offset 16, but is outside bounds of alloc2010 which has size 9
    |
```

I get this error whenever the length of the slice passed to `Arc::copy_from_slice` is not a multiple of 8, and no error when the slice length is divisible by 8.

PR #1097, opened 2020-06-07
copy_from_slice: unsafe, private; pad allocation

divergentdave:
Follow-up to #1096 to fix UB caught with Miri.

The allocation in copy_from_slice needs to be padded out to its
alignment, or else the slice methods called on it may cause undefined
behavior, since the slice's memory layout expects that memory to be
valid.

I also made Arc::copy_from_slice unsafe to call and private, to mirror
the corresponding function in the standard library.


spacejam:
awesome work! I'm really happy about everything miri has caught so far :)

PR #1102, opened 2020-06-14
Removed unused code in TransactionalTree:stage()

pantonov:
Looks like some remnant code, please check...


spacejam:
Thanks! This code is going to be getting a huge overhaul after the current batch of reliability efforts completes

PR #1108, opened 2020-06-22
Conversion of boxed slices to Arc and IVec

### LINKED ISSUE #1107: From<Box<[u8]>> for IVec
This was removed in [0], but I couldn't see it mentioned in the changelog, as it's a breaking change. I was mainly curious why it was removed.

- [0] https://github.com/spacejam/sled/commit/4934a6ba4b5ba7cee8a322b4c7cab6aec136f550#diff-717909ea842788c2cdf7d4ced5665b7c

divergentdave:
This PR rewrites the (unsafe code!) `From<Box<T>> for Arc<T>` implementation to allow dynamically sized types, like `std::sync::Arc` does. Constructing a fat pointer to a dynamically sized `ArcInner<T>` from a `*mut u8` thin pointer and the source fat pointer requires some extra gymnastics, see the comments for details. This implementation is then used to add back the `From<Box<[u8]>> for IVec` implementation. Also included: rewrite of Arc's Debug implementation to allow dynamically sized types, and a small test for the conversion. The unsafe code is modeled after the standard library's code, so it ought to be okay, plus I'm going to run tests with Miri overnight.

Fixes #1107

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -28,9 +28,9 @@ pub struct Arc<T: ?Sized> {
 unsafe impl<T: Send + Sync + ?Sized> Send for Arc<T> {}
 unsafe impl<T: Send + Sync + ?Sized> Sync for Arc<T> {}
 
-impl<T: Debug> Debug for Arc<T> {
+impl<T: Debug + ?Sized> Debug for Arc<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        f.debug_map().entry(&"inner", unsafe { &(*self.ptr).inner }).finish()
+        Debug::fmt(&**self, f)
     }
 }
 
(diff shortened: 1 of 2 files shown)


divergentdave:
Whoops, I goofed up MSRV, will take another look at Layout construction

spacejam:
Has miri thrown any complaints when run against a variety of sizes?

divergentdave:
It looks good. I got a regular old panic when using a ZST, but I fixed that above.

src/arc.rs
```diff
@@ -168,30 +168,49 @@ impl<T: Copy> From<&[T]> for Arc<[T]> {
...
+            // from the source box, and overwrite the pointer to point to the
+            // newly allocated memory. Otherwise, if *mut T is a thin pointer,
+            // this will overwrite the entire pointer, and dst will be
+            // identical to dst_thin, save for its type.
+            let mut dst = src as *mut ArcInner<T>;
+            #[allow(trivial_casts)]
+            ptr::write(
+                &mut dst as *mut _ as *mut *mut u8,
+                dst_thin as *mut u8,
+            );
```
spacejam, line 197: This is the part that I'm stumbling a bit over. I'm a bit worried that this might break if the internal layout for fat pointers changes. The reason I went with the fatten method from https://users.rust-lang.org/t/construct-fat-pointer-to-struct/29198/9 was because @dtolnay provided it as a way to fatten things without assuming the memory layout. I think I want to avoid this if a later version of Rust changes the internal layout for fat pointers and causes this to break. I'm having trouble knowing if this will also break.

divergentdave: Ah, good point. I can make use of that to write a narrower `impl<T> From<Box<[T]>>`.

src/arc.rs
```diff
@@ -168,34 +168,51 @@ impl<T: Copy> From<&[T]> for Arc<[T]> {
...
+    fn from(b: Box<[T]>) -> Arc<[T]> {
+        let len = b.len();
         unsafe {
-            let dst_layout = Layout::new::<ArcInner<T>>();
+            let src = Box::into_raw(b);
+            let value_layout = Layout::for_value(&*src);
+            let align = std::cmp::max(value_layout.align(), mem::align_of::<AtomicUsize>());
+            let rc_width = std::cmp::max(align, mem::size_of::<AtomicUsize>());
+            let unpadded_size = rc_width.checked_add(value_layout.size()).unwrap();
+            let size = (unpadded_size + align - 1) & !(align - 1);
```
spacejam, line 181:
```suggestion
            // pad the total `Arc` allocation size to the alignment of `max(itself, AtomicUsize)`
            let size = (unpadded_size + align - 1) & !(align - 1);
```

spacejam:
Looks good! Thanks for fixing this!

PR #1115, opened 2020-07-03
optimize some memory orderings

spacejam:


spacejam:
This reduces cycles spent in the LRU code by ~30%

PR #1117, opened 2020-07-03
Make Db::drop_tree parameters match open_tree parameters.

hdevalence:
Because the new bound `AsRef<[u8]>` covers the previous type `&[u8]`, I think this is not a breaking change.


spacejam:
@hdevalence thank you :)

spacejam:
I'm expecting to include this change in a 0.32.1 release shortly

PR #1129, opened 2020-07-19
Add IVec::zeroed(len)

### LINKED ISSUE #1122: Feature Request: IVec::zeroed(capacity)
#### Use Case:

I am investigating use of sled with values around 190 KiB in size. When the performance of my multi-threaded protoype fell off a cliff, Valgrind told me why: an astonishing number of machine cycles are spent just copying those buffers around.

Digging into it, I was surprised to learn that `Arc<[u8]>` does not have a zero copy cast from `Vec<u8>`, `Box<[u8]>`, or even `Arc<[u8; N]>`!

As a result, I can see no way to avoid copying data at least once into an `IVec` in the current API. I'd like to have a zero-copy way to populate one.

#### Proposed Change:

I propose adding a new method, `IVec::zeroed(capacity)`. It allows an `IVec` to be created in advance of having the data in a buffer somewhere, if the caller knows what the data size will be (my use case), or can tolerate padding (many others, I'm sure).

Example:

```
let db = sled::open_db(...)?.tree(...);
let generator = ... ;
let expected_len = generator.calculate_len();
let dst = IVec::zereoed(expected_len);
let writer = std::io::Cursor::<&mut [u8]>::new(&mut dst);
assert_eq!(generator.write_into(&mut writer), expected_len);

db.insert("my_data".to_bytes(), dst)?;
```

#### Who Benefits From The Change(s)?

Anyone who wants to use large keys and values.

#### Alternative Approaches

I had several other ideas for alternatives, but I imagined they would be more difficult to implement:

1. Make `IVec` have a separate `capacity` like `Vec<T>`, and support incremental pushes to it.
2. Make `IVec` support arbitrary writers to a growing internal buffer, and then shove that into an `Arc<[u8]>` after the writer is finished.
3. Figure out how to write a zero copy cast from an owned `[u8]` type to `sled::Arc<[u8]>`. (I suspect this hasn't been done, because Rust's layout guidelines would make any pointer tricks UB.)

In addition, having an `IVec::filled(capacity, pattern)` method might be more general for protocols and formats who could not tolerate zeroes.

jhwgh1968:
Closes #1122.

This is a WIP, because the destructors are clearly invoking UB and I don't know why.

Do you have any idea, @divergentdave?

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -3,7 +3,7 @@
 /// `Arc`, but we use a LOT of `Arc`'s, so the extra 8 bytes
 /// turn into a huge overhead.
 use std::{
-    alloc::{alloc, dealloc, Layout},
+    alloc::{alloc, alloc_zeroed, dealloc, Layout},
     convert::TryFrom,
     fmt::{self, Debug},
     mem,
@@ -98,6 +98,33 @@ impl<T> Arc<T> {
         let sub_ptr = (ptr as *const u8).sub(rc_width) as *mut ArcInner<T>;
 
         Arc { ptr: sub_ptr }
+    }
+}
+
+impl Arc<[u8]> {
+    pub fn zeroed(len: usize) -> Arc<[u8]> {
+        let fat_ptr: *const ArcInner<[u8]> = unsafe {
+            let align =
+                std::cmp::max(mem::align_of::<u8>(), mem::align_of::<AtomicUsize>());
+
+            let rc_width = std::cmp::max(align, mem::size_of::<AtomicUsize>());
+            let data_width = mem::size_of::<u8>().checked_mul(len).unwrap();
+
+            let size_unpadded = rc_width.checked_add(data_width).unwrap();
+            // Pad size out to alignment
+            let size_padded = (size_unpadded + align - 1) & !(align - 1);
+
+            let layout = Layout::from_size_align(size_padded, align).unwrap();
+
+            let ptr = alloc_zeroed(layout);
+            assert!(!ptr.is_null(), "failed to allocate Arc");
+
+            #[allow(clippy::cast_ptr_alignment)]
+            ptr::write(ptr as _, AtomicUsize::new(1));
+
+            Arc::fatten(ptr, len)
+        };
+        Arc { ptr: fat_ptr as *mut _ }
     }
 }
 
--- a/src/ivec.rs
+++ b/src/ivec.rs
@@ -107,6 +107,15 @@ impl IVec {
         IVec(inner)
     }
 
+    /// Creates a new `IVec` of a given length filled with zeroes.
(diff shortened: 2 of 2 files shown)


jhwgh1968:
I fixed one clippy issue, but a bunch of others triggered on code I didn't write after some errors.

@spacejam, do you know what the deal is?

divergentdave:
The allocation used to create the `ArcInner` is too small, as it's only sized for the byte array, and doesn't have room for the reference count. Additionally, the reference count should be initialized to one by this constructor. See `Arc::copy_from_slice` and the `From<Box<[T]>>` implementation for examples.

jhwgh1968:
Thanks for the pointers (no pun intended) @divergentdave! It's fixed now.

spacejam:
This is high-risk code to accept the maintenance burden of over time. It's not clear that this actually improves your performance (and what do you mean by performance, are you talking about latency, throughput, allocations, etc...?)

jhwgh1968:
I thought you were okay with that, @spacejam, given your custom Arc re-implementation. But I can certainly provide more clarity on my motivation, rather than just "zero copy seems like a good idea".

It will take me a couple days to double-check my graphs, and come up with a public reproducer for your examination to prove this method fixes the issue. But here is the gist:

I have a Rayon thread pool running jobs in an `into_par_iter` construction. Each job will do significant number crunching from a small input,  produce 190 KiB of output, and write it to a unique key in a sled tree shared between all the threads.

This is similar to other workloads I have thrown at sled, though the chances of CPU starvation is higher. But if sled is starved for short to medium periods of time, I don't mind.

What I didn't expect was the program to slow down more and more, the larger the job was. Valgrind told me it was because of `memcpy` . So I did a massif profile, and found that my heap was growing by leaps and bounds -- 50-100% each time the thread pool split. My largest job which stalled for minutes had a heap peak of over 3 GiB.

Why am I pointing the finger at sled rather than rayon? Three things:

1. According to Massif, the incremental change in heap was attributed to one of the thread pool inner functions doing the split. But they were always calling `sled::Arc::from_slice` at the time of the peaks. That is, the copy from a `Vec` into the `Arc` inside of an `IVec` was the tip of the peak, for every peak.

2. The "mountain shape" of the heap usage -- with a continuous slope, and then a steep drop off near the end when the jobs were done -- strikes me as a queue backing up and then catching up. Doesn't sled write periodically, and wouldn't that thread back up if starved?

3. When I removed sled -- having each thread dump the output into one temporary file each -- this exponential heap growth went away.

If you'd prefer I make a separate issue for this, I'd be happy to.

jhwgh1968:
After doing a more thorough test, it seems that using `IVec:zeroed` did not have the effect I expected. The heap growth is still there from other parts of `sled`.

If I can more precisely characterize the behavior, I will open a separate bug. In the mean time, I will close this PR.

PR #1157, opened 2020-08-29
Make IVec writable and sequentially readable (extend_from_slice, io::Write, io::Read, etc.)

pantonov:
- Implemented Vec-like methods `new()`, `with_capacity()`, `capacity()`, `extend_from_slice()`, `push()`, with COW semantics; this was easy because IVec uses shared Arc for buffer and already used COW for subslicing
- Implemented io::Write, io::Read traits for IVec (read works via subslicing)

With these changes, IVec becomes much more convenient to use with I/O methods,. serialization libraries like serde etc.

Internally, some minor refactoring and few utility methods for Arc<[u8]>, also added T: Copy trait bound for conversion trait impls for Arc owning slices, otherwise it may be unsound (e.g. Vec<T> -> Arc<[T]> uses bitwise copying).

### DIFF
--- a/src/arc.rs
+++ b/src/arc.rs
@@ -41,14 +41,16 @@ impl<T> Arc<T> {
         Arc { ptr }
     }
 
-    // See std::sync::arc::Arc::copy_from_slice,
-    // "Unsafe because the caller must either take ownership or bind `T: Copy`"
-    unsafe fn copy_from_slice(s: &[T]) -> Arc<[T]> {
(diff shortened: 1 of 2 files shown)


spacejam:
Hey @pantonov, thank you for the effort you spent on this, but I think it would be a mistake for us to add this big chunk of fairly unsafe logic to `IVec`, which I have been working to simplify more and more over time because it was getting out of control.

Issue #1152, opened 2020-08-22
Transactions are deadlocking at insert_inner


D1plo1d:
Transactions are deadlocking inside of commit and never returning in my application. LLDB shows the thread as stuck on this line:

```rust
while self.tree.insert_inner(k, v_opt.clone(), &mut guard)?.is_err()
{
}
```

Source things: https://github.com/spacejam/sled/blob/master/src/transaction.rs#L364

1. **expected result:** I expect `TransactionTree::commit` to always return a result.
1. **actual result:** `TransactionTree::commit` intermittently blocks until I kill the application.
1. **sled version:** `0.34.2` + some small error handling experiments: https://github.com/D1plo1d/sled/tree/53d0e30a749f27fb808101730f1794a5f85b6216
1. **rustc version**: `rustc 1.44.1 (c7087fe00 2020-06-17)`
1. **operating system**: Ubuntu 20.04
1. **minimal code sample that helps to reproduce the issue**: Intermittent issue. TBH I could use a hand figuring out how to reliably reproduce my issue.
1. **logs, panic messages, stack traces**: Not sure what would be helpful here (please let me know!). I added a couple `trace!` macros to commit and observed that the commit started but never finished.

spacejam:
Could you please supply the additional information from the bug template, to help track down what may have happened?

D1plo1d:
Will do - sorry for the delay I think I'll have time to give this a proper look into this this week.

D1plo1d:
@spacejam: I've filled in what I have. If there's more that I can do to help let please me know.

I have a note here that the deadlock was shown by LLDB to be in the line `sender.send(rx).is_err() {` in subscriber.rs and that in turn calls `self.mu.lock();` in threadpool.rs. Is there anything that could cause the `threadpool::Queue` to permanently lock?

spacejam:
@D1plo1d thanks for the additional info! That should help to track this down quickly :)

D1plo1d:
@spacejam Appreciate it! I'm giving this my best unqualified debugging attempt XD

> Condvar will always perform an unfair unlock when releasing the mutex before a wait.
Source: https://github.com/Amanieu/parking_lot/issues/238

I stumbled upon that in Parkinglot's issues. Could Condvar's wait unfairness be related to this?

spacejam:
[...]

PR #1159, opened 2020-08-31
debug subscription + transaction deadlock

spacejam:
this is the ongoing workspace for fixing the bug found in #1152

### DIFF
--- a/src/oneshot.rs
+++ b/src/oneshot.rs
@@ -2,6 +2,7 @@ use std::{
     future::Future,
     pin::Pin,
     task::{Context, Poll, Waker},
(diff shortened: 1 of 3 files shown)


soro:
I have to say that I can't find another obvious deadlock that would lead to an issue in this particular scenario - the interface is such though that it is pretty easy to deadlock if the thread processing the subscription does anything else involving subscriptions, such as try to subscribe to another prefix. Maybe it would be best just to never block on the sending side and just use try_send? That or simply use a concurrent map as the registry so those locks don't have to be taken while sending is going on.

PR #1161, opened 2020-09-03
fix destructor deadlock from issue #1152

soro:
This should fix the deadlock from spacejam/sled#1152

https://github.com/spacejam/sled/issues/1152#issuecomment-686390978

### DIFF
--- a/src/subscriber.rs
+++ b/src/subscriber.rs
@@ -1,6 +1,7 @@
 use std::{
     future::Future,
     pin::Pin,
+    mem::ManuallyDrop,
     sync::{
         atomic::{AtomicBool, Ordering::Relaxed},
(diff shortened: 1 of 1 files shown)


spacejam:
LGTM, thanks!

spacejam:
hmm, when merged in with the tyler_subscriber_send_lock branch with the test for this, it still locks up sometimes

spacejam:
similar stacktraces, with a writing thread blocked on the `sender.send` call in subscriber.rs. Going to revert this for now because it doesn't seem to address the test that is locking up.

soro:
Interesting. It definitely is a scenario that could currently lead to deadlock, so I think it needs to be added either way eventually. But that means there is something else as well. I will look into it some more.

PR #1164, opened 2020-09-05
Widen trait bounds to allow Into

### LINKED ISSUE #999: Minor API proposal: change `From` to `Into`

Plecra:
This closes #999

All the changes were mechanical, so I wouldn't expect there to be any issue.

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -45,16 +45,17 @@ impl Batch {
     /// Set a key to a new value
     pub fn insert<K, V>(&mut self, key: K, value: V)
     where
-        IVec: From<K> + From<V>,
+        K: Into<IVec>,
+        V: Into<IVec>,
     {
-        self.writes.insert(IVec::from(key), Some(IVec::from(value)));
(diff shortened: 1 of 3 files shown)


spacejam:
Thanks!

PR #1169, opened 2020-09-10
debug free pid panic

spacejam:


spacejam:
This happens because nodes are sometimes merged by another thread while fmt is running

PR #1178, opened 2020-09-22
[breaking batch] add Event::Batch that receives atomic batches from transactions and batch writes. Closes #1165.

### LINKED ISSUE #1165: Subscription events are not consistent with the order of operations inside transactions
**Note to the maintainers:** I'm hoping this is helpful in determine where to take the implementation next but please let me know if it's becoming too much about subscriptions at once.

This is two subscription issues that are both cause by the TransactionalTree `writes` HashMap's arbitrary iteration order and the fact that it only stores the final IVVec value for each key:

#### Scenario:

1. watch_prefix `"thing"`
1. start a transaction
1. insert at key `"thing_1"`
1. insert at key `"thing_2"`
1. commit

#### Outcome

As I understand it the subscription will now have an arbitrary event order for the inserts of thing_1 and thing_2.

#### Preferred Outcome

It would be useful to me as a developer to be able to modify two things in a specific order with the atomic guarantees of a transaction and have those effects come out of the subscription in the same order as well.

For example, if I wish to add two items to a single threaded work queue such that thing_1 gets executed before thing_2 I cannot do this atomically at the moment.

The work queue could of course be rewritten to be entirely stored in a single key but this looses some benefit of subscriptions as I have to then write my own logic to determine which item(s) changed and also the work queue would risk becoming arbitrarily large and slow to deserialize.

#### Scenario:

1. watch_prefix `"thing"`
1. start a transaction
1. insert at key `"thing_1"`
1. insert at key `"thing_1"`
1. commit

#### Outcome 

This one I expect to be controversial but I wanted to at least start the conversation. If I make two changes to a single key in a transaction I *think* I'd expect to see two events in the subscription as if they were made non-transactionally (serialized?) however at present AFAIK the subscription will AFAIK only ever see the final state.

This second scenario is more of an observation that might not be congruent with expectations - I was at least surprised to see it but I'm not 100% sure how this plays in with transaction atomicity or if I'd prefer something else. What do y'all think - what would be your expected behaviour here?

spacejam:


spacejam:
cc @D1plo1d

spacejam:
I've restructured Event to be unified with batches and multi-tree transactions. While iterating over &Event, you now also get a reference to the Tree that specific updates happened on.

PR #1199, opened 2020-11-26
implement `From<&String>` for `IVec`

nooberfsh:


spacejam:
Hey @nooberfsh, thanks for the PR! While `From<&str>` is already implemented, I suppose that this might reduce friction a little bit. I'm going to also add `From<String>` to further reduce friction for folks.

Issue #1124, opened 2020-07-13
page heap


spacejam:
after loading the dict.cc DE->EN dictionary, 66mb, into sled, sled uses 144mb (2.18 space amplification) without compression enabled.

Here's the percentiles of used space per storage segment:
(512k is a full segment)
```
storage 0: 0
storage 5: 84964.45150791103
storage 10: 92040.97481768382
storage 15: 95797.27906818989
storage 20: 98714.7710107605
storage 25: 100708.96186695219
storage 30: 103776.0368200868
storage 35: 105872.47202785847
storage 40: 108011.25803610642
storage 45: 110193.25038775145
storage 50: 111300.72101586298
storage 55: 113549.16492766769
storage 60: 118183.23506803908
storage 65: 121782.48086768923
storage 70: 129313.15075081984
storage 75: 153275.69022931982
storage 80: 358612.3257686633
storage 85: 373247.61322898994
storage 90: 388480.17810109997
storage 95: 392384.4788500711
storage 100: 429337.10025710246
segments free 3 total 309
```

It's clear that by mixing our update log with rewritten pages, it causes a ton of fragmentation. We need to separate delta updates from rewritten/merged pages. We can initially write pointers into the log, as we do with blobs. Then when we perform the next fuzzy checkpoint that merges the pointer into the snapshot, it can be forgotten by the log.

There is a trade-off between RCU in the heap vs aries-style logging of undo operations along with redo. Under the hood, flash storage has to rewrite 4k pages anyway, and flash performance suffers when there are non-contiguous ranges of a file that get mapped to different places, incurring more indexing complexity by the FTL.

We can start with a simple 64k fixed size heap, and keep using the blob approach for anything that overflows. 

* On-log page rewrites go straight to the heap
* CAS'd pages go straight to the heap

to write things to the heap:
1. a slot is reserved and written to
2. a pointer is written to the log
3. a peg is added to the stability tracker, as batches are currently tracked
4. when the heap is fsynced, all pending pegs are cleared in the stability tracker

over time, the pointer is rewritten, the same way blob pointers are rewritten, but we simply drop any pointers that are older than the beginning range of the last stable fuzzy checkpoint.

to further reduce space amplification, we can either use multiple heaps for various page sizes, and/or use hole punching on page slots that are vacated.

spacejam:
Closed in #1214. Additional tuning needs to happen to reduce space further, but initial results are quite promising.

Issue #1209, opened 2020-12-08
shred of page pointer detected in crash test on master branch


spacejam:
```
     Running target\release\deps\test_crash_recovery-0846fd0898adb97e.exe
thread 'main' panicked at 'we shredded the location for pid 2
                        with locations Present { base: (39616, Inline(39616), 38), frags: [(39695, Inline(39695), 21), (39716, Inline(39716), 17), (39774, Inline(39774), 21), (39795, Inline(39795), 17), (39853, Inline(39853), 21), (39874, Inline(39874), 17)] }
                        by zeroing the file tip after lid 39829', src\pagecache\snapshot.rs:430:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'crash_tx test child exited abnormally', tests/test_crash_recovery.rs:295:9
```

spacejam:
Happened again, both times have been on windows

```
thread 'main' panicked at 'we shredded the location for pid 2
                        with locations Present { base: (19476, Inline(19476), 38), frags: [(19555, Inline(19555), 21), (19576, Inline(19576), 17), (19634, Inline(19634), 21), (19655, Inline(19655), 17)] }
                        by zeroing the file tip after lid 19610', src\pagecache\snapshot.rs:430:21
```

PR #1214, opened 2020-12-14
slab style heap storage engine

spacejam:


spacejam:
the remaining test flakes are #1209

spacejam:
Closes #1124.

PR #1221, opened 2020-12-19
Add filtering and stderr output to no-harness test

divergentdave:
This PR improves the ergonomics of the test_crash_recovery test. Since this integration test has `harness = false` set, it is up to us to handle test filtering, console output, etc. These changes add console output that approximates what we see from the standard library test harness, and allow filtering out tests using the same command line argument interface. Sample output:

```
david@desktop:~/Code/sled$ cargo test --features=testing -- crash_tx
    Finished test [unoptimized + debuginfo] target(s) in 0.05s
     Running target/debug/deps/sled-e06fd382104117c8

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 29 filtered out

     Running target/debug/deps/test_crash_recovery-83cd0a124e744f4b

running 1 test
test crash_tx ... ok

test result: ok. 1 passed; 3 filtered out

     Running target/debug/deps/test_log-440f4f40cf3c48f3

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 7 filtered out
...
```

### DIFF
--- a/tests/test_crash_recovery.rs
+++ b/tests/test_crash_recovery.rs
@@ -26,6 +26,13 @@ const BATCHES_DIR: &str = "crash_batches";
 const ITER_DIR: &str = "crash_iter";
 const TX_DIR: &str = "crash_tx";
 
+const TESTS: [(&'static str, fn()); 4] = [
+    (RECOVERY_DIR, crash_recovery),
+    (BATCHES_DIR, crash_batches),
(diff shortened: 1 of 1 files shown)


spacejam:
This is great, I can't tell you how many times I've manually commented out the crash tests to prevent them from running while trying to run a particular regression test :]

PR #1222, opened 2020-12-20
Don't try to call fallocate in Miri

divergentdave:
`libc::fallocate` does not have a shim implemented in Miri, this small patch conditionally disables hole punching in Miri to avoid calling it.


spacejam:
Thanks!

Issue #1077, opened 2020-05-23
ALICE-like write permuter


spacejam:
alice (#83) is a tool that uses strace to record file modifications, then uses models of various file systems to permute copies of the output file, and then tries to load the file back into the system to assert basic correctness properties of the system. this test methodology found bugs in every system it targeted.

we can build a lightweight version by just recording file writes and permuting the file in various ways. we can record a trace of file operations, represented as fsyncs, writes, truncate operations, etc...

once we have the trace, we can "replay" the operations to construct various log files that we try to load and assert various properties about. when replaying the operations, we can reorder any operations that happen between fsyncs. we can start with an empty file and grow it forward, trying different permutations of present items depending on where writes happened in relation to syncs. there tend not to be so many operations between sync calls, so our histories are actually not so expensive to exhaustively enumerate.

interesting bugs that are specific to sled's guarantees can also be tested for. because sled rewrites old log files without zeroing them first (we zero on recovery to prevent "replay" attack-like corruption). we should be able to write arbitrary subsets of data, as well as intentionally flipping bits between fsyncs, and still maintain our desired properties. sled can sort of be interpreted as a heuristic for recovering partially corrupted files, anyway.

spacejam:
It's preferable to use our version compared to the actual ALICE tool because the actual tool is pretty heavy-weight, and by instrumenting our code directly, the overall complexity can be lower and performance can be much higher.

divergentdave:
This is an exciting avenue! One thing I like is that it could also catch some of the issues the failpoints and crash recovery tests do, but reproducibly, by chopping off the trace at different points. (maybe once the write permuter is stood up, we could make a hybrid ALICE/failpoints test that would allow reproducing trickier things like double-recovery bugs)

I think the best ways to collect traces would be to either add a bunch of IO operations to the existing event log infrastructure, (perhaps behind yet another feature flag?) or to move all IO operations behind a trait, make implementations of the trait for real IO, traced IO, and mocked IO, and parameterize the PageCache on that trait. I'm not sure which would be the best route; my chief concerns are the cost of copying so many buffers for the event log affecting non-ALICE tests, and the possible developer UX impacts of introducing a generic type parameter on public types, even if it has a default.

spacejam:
[...]

PR #1229, opened 2020-12-22
ALICE fixes

divergentdave:
See also #1077. This branch holds the changes I've made thus far in response to issues found by ALICE. The entire test setup can be found at https://github.com/divergentdave/alice-sled, including a git submodule pointing to this branch.

Currently I'm not getting any more reports from the tool, though I haven't waited long enough for it to terminate yet. I would like to expand the workload and checker to exercise more functionality. For example, I expect we'll at least need more fsyncs in the heap directory.

Big picture, it would be good to turn each crash recovery test into a workload/checker pair, where the workload somehow serializes the equivalent of the test's reference data structure to stdout, and the checker reads stdout to reconstruct the reference data structure, and verify if the particular durability invariant was maintained in the database.


divergentdave:
It appears that Windows does not have an equivalent to FlushFileBuffers for directories. We'll have to conditionally skip the directory sync.

src/config.rs
```diff
@@ -558,7 +558,9 @@ impl Config {
             self.get_path().join("DO_NOT_USE_THIS_DIRECTORY_FOR_ANYTHING"),
         );
 
-        self.try_lock(options.open(&self.db_path())?)
+        let file = self.try_lock(options.open(&self.db_path())?)?;
+        fs::File::open(self.get_path())?.sync_all()?;
+        Ok(file)
```
spacejam, line 563: Ahh interesting, this makes a lot of sense in hindsight, but I don't think I've ever thought to fsync a file that I've just opened before!

divergentdave: This fsync is actually on the directory containing the file, so it doesn't disappear. I'll move this into a function call, so it will be more clear.

divergentdave:
I have ported the insert and batch crash recovery tests to ALICE workloads so far, and I just turned up another issue running the batches workload. (see https://github.com/divergentdave/alice-sled) It appears we need to fsync the snapshot directory after renaming the file to its final place, and before writing to the pagecache in a way that depends on the snapshot's presence. I'll add this fsync to my branch and see if that does it.

```
ALICE output:

...
130	write("db", offset=10910, count=353, inode=10921119)
131	fsync("db", size=11264, inode=10921119)
132	creat("snap.0000000000002E9E.generating", parent=10921111, mode='0666', inode=10915824)
133	append("snap.0000000000002E9E.generating", offset=0, count=273, inode=10915824)
134	append("snap.0000000000002E9E.generating", offset=273, count=8, inode=10915824)
135	append("snap.0000000000002E9E.generating", offset=281, count=4, inode=10915824)
136	fsync("snap.0000000000002E9E.generating", size=285, inode=10915824)
137	rename(dest="snap.0000000000002E9E", source="snap.0000000000002E9E.generating", source_hardlinks=1, source_parent=10921111, dest_size=0, dest_hardlinks=0, source_size=285, dest_parent=10921111, source_inode=10915824, dest_inode=False)
138	write("heap/00", offset=0, count=65536, inode=10921122)
139	write("heap/00", offset=65536, count=65536, inode=10921122)
140	write("heap/00", offset=131072, count=65536, inode=10921122)
141	write("heap/00", offset=196608, count=65536, inode=10921122)
142	write("db", offset=5120, count=1024, inode=10921119)
143	write("db", offset=8192, count=1024, inode=10921119)
144	fsync("db", size=11264, inode=10921119)
...

(Dynamic vulnerability) Ordering: Operation 137 needs to be persisted before 143

thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `5`,
 right: `3`: expected key 6 to have value 5, instead it had value 3', cases/batches/src/lib.rs:23:9
```

divergentdave:
I turned the `test_tree_failpoints` test into a fairly complicated workload, with crash recovery and more fine-grained information for use in the checker. The one issue I've found so far is that if we generate an ID, then crash and recover when no flushes have taken place since startup, we can generate an ID of 0 a second time, because the "meta" and "counter" pages haven't made it to the disk yet, and `PageCache::start()` sets `was_recovered = false` and stores 0 in `idgen`. I fixed this by making `PageCache::start()` block on flushing these two pages if the database is brand new.

divergentdave:
The next issue I see is that we can generate an ID a second time after recovering from a crash, and the doubled ID seems to always be a multiple of two times idgen_persist_interval. I haven't found a root cause for this one yet, I'll need to read through the relevant code first.

divergentdave:
I figured it out, the first blocking flush of the counter needs to happen sooner after recovery. The call to make_durable() in generate_id_inner() will only block until the prior counter page update is flushed, not the most recent one. When recovering a database, we add two intervals to `idgen`, which effectively blows through the buffer that we rely on when justifying flushing the past update. The first time generate_id_inner() is called after recovery, it will always update the counter page, the LSN of the counter page will necessarily be stable, since we just read it during recovery, and then make_durable() will always return without blocking. This gives us a window where an ID can be returned with no disk writes after recovery. In a pathological scenario, we could imagine starting the database, getting one ID, and immediately crashing in a loop such that the same ID is returned every time.

I observed this in an ALICE trace that showed the following:
* db.restart(), followed by related file operations
* db.generate_id() called
* ID of 3 returned
* db.generate_id() called
* write of 18 bytes to db
* fsync of db
* ID of 4 returned

I reviewed the 18 byte write above, and it was two messages, `Update::Counter(3)` and `Update::Counter(4)`.

I think the best way to resolve this would be to CAS in an updated counter page during PageCache::start(), with the two intervals added. In the happy case, this will ride along with another IO buffer write before generate_id() is called, and in the worst case, the same logic in generate_id() will now block until the updated counter from recovery is written, because its LSN is now the 'previous LSN' that we pass to make_stable().

src/pagecache/mod.rs
```diff
@@ -719,17 +719,34 @@ impl PageCache {
...
+                // will advance them further by another interval, and wait for
+                // this update to be durable before returning the first ID.
+                let necessary_persists = (counter / pc.config.idgen_persist_interval + 1)
+                    * pc.config.idgen_persist_interval;
+                let counter_update = Update::Counter(necessary_persists);
+                let old = pc.idgen_persists.swap(necessary_persists, Release);
+                assert_eq!(old, idgen_persists);
+                // CAS should never fail because the PageCache is still being constructed.
+                pc.cas_page(COUNTER_PID, idgen_key, counter_update, false, &guard)?.unwrap();
+            }
```
spacejam, line 749: I agree with your reasoning and this seems like a nice fix!

spacejam:
This is great work! It makes me so happy to see your work with ALICE yielding such high quality insights into correctness, which are able to be addressed with such clean diffs :)

crusty-dave:
Where is the ALICE tool?  I can only seem to find a very stale reference here:
https://research.cs.wisc.edu/adsl/Software/alice/doc/adsl-doc.html

Thanks in advance,
-Dave

PR #1230, opened 2020-12-26
add IVec::serialize/view_as

### LINKED ISSUE #1064: IVec::view_as<'de, T: serde::Deserialize<'de>> method
#### Use Case:

users frequently want to access high-level structures rather than raw bytes

users also may use multiple types of structures in the same `sled::Tree` keyspace.

most users probably already use `serde`.

#### Proposed Change:

add a method to `IVec` for attempting to view the data as a high-level struct that implements `serde::Deserialize` without copying underlying data.

#### Who Benefits From The Change(s)?

users who wish to use high-level structures with sled

#### Alternative Approaches

zerocopy instead of serde. This has the disadvantage that people are not familiar with zerocopy.

### LINKED ISSUE #1065: impl From<T: serde::Serialize> for IVec
#### Use Case:

users often use serde for serializing data into sled

#### Proposed Change:

`impl From<T: serde::Serialize> for IVec` so users can use high-level structures directly without messing with serde, replacing the other `From` impls for `IVec`.

#### Who Benefits From The Change(s)?

users using high-level structures with sled

naim94a:
I think this project is really cool!
This should close #1064 and close #1065. 

- Will optionally pull-in serde as "c_serde" if the "serde" feature is enabled.
- It's not possible to `impl From<T: serde::Serialize> for &T` due to conflicting implementations, so I named the function `serialize`.
- uses `src/serialize.rs` to serialize primitives.

Should this API save field names & types (like a binary version of JSON/msgpack)?  The current implementation doesn't retain type information or field names. Structures serialized with this API must use the same structure to deserialize for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ io_uring = ["rio"]
 docs = []
 miri_optimizations = []
 mutex = []
+serde = ["c_serde"]
 
 [dependencies]
 crossbeam-epoch = "0.9.1"
@@ -52,6 +53,7 @@ parking_lot = "0.11.1"
 color-backtrace = { version = "0.5.0", optional = true }
 rio = { version = "0.9.4", optional = true }
 backtrace = { version = "0.3.55", optional = true }
+c_serde = { package = "serde", version = "1.0", optional = true}
 
 [target.'cfg(any(target_os = "linux", target_os = "macos", target_os="windows"))'.dependencies]
 fs2 = "0.4.3"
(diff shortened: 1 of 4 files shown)


spacejam:
Hey @naim94a, thank you for the effort you spent on this, but the ideas for this are still quite in-flux, as they have significant implications for how tied we become to other libraries. I think that the approach sketched out in #1266 might be a better fit, because it would not peg us to any specific serialization library, but we still need to discuss a lot of details before moving ahead to implementation.

Issue #1219, opened 2020-12-18
inline index nodes with eytzinger layouts


spacejam:
![image](https://user-images.githubusercontent.com/1505488/102588875-6a146080-410e-11eb-9016-c3196a8dd401.png)
https://arxiv.org/pdf/1509.05053.pdf

We spend a lot of time doing binary search in index nodes to find an interesting leaf. This is partially because we don't do interpolation search, but mostly because we are very cache inefficient. Index keys tend to be very short due to the combination of prefix compression and suffix truncation, and the current IVec implementation that inlines up to 22 bytes wastes a ton of space for these short items. A path forward is to inline everything, and use an eytzinger layout. We don't want to use this for leaf nodes too because of the cost it imposes on sequential workloads, but it's probably a great choice for index nodes that point to the leaves.

PR #1231, opened 2020-12-28
single-allocation nodes

spacejam:
this large update changes how tree nodes are represented:
* a single contiguous allocation for a node's metadata, keys, and values
* variable offset and length bytes
* avoids offset tables when keys and values are fixed length
* combines offset tables when keys and values are both variable length
* zero additional serialization / deserialization effort
* likely addresses cache sizing problems

over time, this makes it easy to take advantage of
* eytzinger layouts #1219 
* cheaper cache accounting
* far more compact in-memory representations
* essentially free serialization

# todo

- [x] reimplement prefix encoding
- [x] reimplement suffix truncation
- [ ] Arc<[u8]> with views to facilitate zero-copy reads
- [x] compare cpu profiles
- [x] compare memory profiles, assert cache size is better respected now
- [ ] optimize delta application

### DIFF
--- a/benchmarks/stress2/Cargo.toml
+++ b/benchmarks/stress2/Cargo.toml
@@ -29,6 +29,7 @@ env_logger = { version = "0.7.1", optional = true }
 log = { version = "0.4.8", optional = true }
 color-backtrace = { version = "0.3.0", optional = true }
 jemallocator = { version = "0.3.2", optional = true }
+num-format = "0.4.0"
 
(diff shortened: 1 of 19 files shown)


spacejam:
Initial heap profiling with massif is showing that this change seems to cut overall memory usage by at least 25% :) 

More tuning, especially around node aging and the interaction with fuzzy snapshots and the new heap, will happen in subsequent PRs.

PR #1233, opened 2021-01-01
[WIP] Use `fetch_max` instead of `bump_atomic_lsn`

Yam76:
The relevant part of issue rust-lang/rust#48655 was resolved in rust-lang/rust#72324 so this PR should eventually fix the corresponding TODO in `pagecache`.

This initial PR does some of the grind work of replacing `bump_atomic_lsn`.

**NOTE**: I'm not familiar with Rust's atomics, so I don't know if SeqCst is actually what you want.

For `cargo test`, my OS (Windows 10, WSL) couldn't hole punch the heap, and I got a lot of 60 second timeouts. Also, the tests got stuck on my end at around `tree_bug_39`, so I had to kill the process. It might be better on your end.

### DIFF
--- a/SAFETY.md
+++ b/SAFETY.md
@@ -95,6 +95,7 @@ durability model
       * return None if last_lsn_in_batch >= self.max_lsn
     * batch requirement set to last reservation base + inline len - 1
       * reserve bumps
+	  	* TODO: Change this
         * bump_atomic_lsn(&self.iobufs.max_reserved_lsn, reservation_lsn + inline_buf_len as Lsn - 1);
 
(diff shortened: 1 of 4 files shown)


spacejam:
Hey @Yam76, this is explicitly avoiding `fetch_max` which did not stabilize until Rust `1.45`. sled is currently targeting `1.39` as the minimum supported rust version (MSRV).

Yam76:
D'oh, that's what I get for not reading the full README. Thanks for the response.

spacejam:
Thanks for putting in the effort anyway! :)

Issue #1235, opened 2021-01-04
Missing documentation about what happens to `f` in `sled::Db::transaction`.


vi:
[`sled::Db::transaction`](https://docs.rs/sled/0.34.6/sled/struct.Db.html#method.transaction) accepts a `Fn` instead of `FnOnce`.

This suggests that Sled may run the transaction function more than once, i.e. automatically retry transactions in case of conflicts.

The documentation of the function however does not mention whether transactions are retried automatically and in general what happens in case of conflict. I think it should be specified explicitly, along with recommendation to avoid using interiour mutability, to avoid heavy calculations, etc. (if what am I understanding is correct).

Additionally the latter part of the documentation suggests to use a pair of `sled::Tree`s as a queue. Users may adapt that pattern into their projects, modifying it to perform some heavy and/or IO-intensive computation on the way from `tx_unprocessed` to `tx_processed`, then introduce some parallelism. If transactions work with compare-and-swap-style retries then such design would be clearly inferiour. I think those matters should also be discussed in the document. 

1. expected result: Conflict handling and design recommendations are described or linked in `sled::Db::transaction` doccomment.
1. actual result: No such info in that doc.
1. sled version: 0.34.6

spacejam:
Please feel free to open a PR

stevenroose:
So this was not really resolved, right? Does it really have to be `Fn` instead of `FnMut`?

It makes it really annoying to move variables into the closure if it's `Fn`. Is the closure really used more than once?

stevenroose:
Created this issue for some explanation: https://github.com/spacejam/sled/issues/1450.

Issue #1234, opened 2021-01-04
Sled database is not really closed after dropping `sled::Db`.


vi:
Example code snippet (non-standalone):

```rust
        let db: sled::Db = sled::open(&opts.database_path)?;
        db.insert(b"qqq2", b"www3")?;
        db.flush()?;
        drop(db);


        let db: sled::Db = sled::open(&opts.database_path)?;
        drop(db);
```

1. expected result: always works. No extra threads or file descriptors or other remnants remain after dropping all `sled::Db` instances.
1. actual result: sometimes fails to close the file descriptor on `drop`, then fails `flock` on the next open. 
1. sled version: 0.34.6
1. rustc version: rustc 1.49.0-nightly (8dae8cdcc 2020-10-12)
1. operating system: Linux

spacejam:
Is this a sled issue or does this have to do with normal linux fd reaping behavior?

vi:
What do you mean by "fd reaping"? Releasing flock on descriptor close or closing descriptors on process exit?

vi:
Does Sled use any global singetons under the hood by the way?

spacejam:
There is a choice about whether to fail if the advisory lock cannot be taken immediately, or to block indefinitely. When testing sled, we use the blocking approach because fd reaping is a lazy process and can take a few moments, and under test we know that we are the only ones using a database. But when running outside of test mode, we do not block indefinitely, because that would provide the user with no actionable information about what the problem is. The code you've listed in the example is expected to fail often, because reaping is lazy. Please keep the database open while using it, and avoid using it from multiple processes. The lazy reaping should almost always happen before you can get around to opening the database again in a new process.

vi:
For example, how would a correct code for closing the database, then causing underlying filesystem unmount (and storage eject) would look like?

divergentdave:
Sled closes the file before drop returns. The FD reaping process @spacejam refers to is a behavior of the Linux kernel. There isn't anything else sled can do during shutdown to affect this.

I have had need to repeatedly close and reopen a database repeatedly for crash testing, and this can be done with a blocking flock() call on the file before opening the database. (via fs2's FileExt::lock_exclusive()) However, I would not recommend this for normal use. Unnecessarily closing and reopening a database is expensive, as each time it will perform database recovery, reading the snapshot/log to load metadata.

vi:
> Sled closes the file before drop returns.

If so, `flock`'s lock gets released automatically and further open should succeed.

```
[...]

Issue #1154, opened 2020-08-25
Question: Is Sled multi-process safe?


ckaran:
I have access to a cluster system, but the nodes on the cluster can only synchronize through the filesystem.  [`flock`](https://man7.org/linux/man-pages/man2/flock.2.html) does work on the system, so I've been using it to ensure that any shared files aren't corrupted, and I could continue to do the same when using sled, but if sled is already multi-process safe, then there's no need for all this code (and it would likely slow sled down a lot).  So, is sled multi-process safe?

spacejam:
sled uses advisory locks, but I haven't subjected them to particularly intense concurrency checking, and I know that in general this is an aspect of filesystems that sometimes is buggy. Importantly, you have to start sled any time you want to use it. It's not possible as of now for multiple processes to share the same open file, and data loss will happen if you do something that lets multiple sled processes work on the same file while bypassing the internal locking logic somehow.

ckaran:
Got it, thank you!  And I think you've managed to address a separate question that's bugged me on and off for years, that is 'why do databases insist on using sockets when the filesystem could be faster'.  Now I know!

itamarst:
I gather from the comment above that the answer is "no, it's not multi-process safe". I had the same question and couldn't find the answer in the documentation, so it might be nice to document this.

visig9:
I think the answer is "sled cannot run in multi-process situation, but If the OS / FS support advisory lock, and implement correctly, datafile would not be corrupted". So it's safe when non-intended run multiple process in the sametime. If I'm wrong please correct me.

Check here:

https://github.com/spacejam/sled/blob/v0.34.6/src/config.rs#L580-L608

ckaran:
@visig9 I would be very careful in relying on this, because as @spacejam said earlier, "...I haven't subjected them to particularly intense concurrency checking, and I know that in general this is an aspect of filesystems that sometimes is buggy."  I know that one of the goals of sled is to not wake up operators in the middle of the night, which means a *lot* of effort has been put into making it operate correctly *if you use it as intended*.  Sled is not intended to be run in the way that this question originally asked (two wholly separate processes opening up the same database at the same time), so there is less testing being done along this path.

[...]

PR #1237, opened 2021-01-05
Extend some doccomments based on Github discussions

vi:
Addresses:

* #1234
* #1235
* #1154

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -5,9 +5,6 @@ use crate::*;
 /// The `sled` embedded database! Implements
 /// `Deref<Target = sled::Tree>` to refer to
 /// a default keyspace / namespace / bucket.
-///
-/// Note that dropping `Db` does not gurantee closing the database. 
-/// Sled may still access the underlying directory under the hood.
 #[derive(Clone)]
 pub struct Db {
     #[doc(hidden)]
--- a/src/db.rs
+++ b/src/db.rs
@@ -5,6 +5,9 @@ use crate::*;
 /// The `sled` embedded database! Implements
 /// `Deref<Target = sled::Tree>` to refer to
 /// a default keyspace / namespace / bucket.
+///
+/// Note that dropping `Db` does not gurantee closing the database. 
+/// Sled may still access the underlying directory under the hood.
 #[derive(Clone)]
 pub struct Db {
     #[doc(hidden)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,11 +8,6 @@
 //! are supported with the
 //! [`Db::open_tree`](struct.Db.html#method.open_tree) method.
 //!
-//! Sled is not designed to be used by multiple processes simultaneously.
-//! Although Sled tries to use advisory filesystem locking to
-//! protect the database from accidental access from foreign
-//! processes, this should not be reiled upon.
-//!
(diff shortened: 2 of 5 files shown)


src/tree.rs
```diff
@@ -335,6 +337,9 @@ impl Tree {
     /// assert_eq!(&processed.get(b"k3").unwrap().unwrap(), b"yappin' ligers");
     /// # Ok(()) }
     /// ```
+    ///
+    /// Note simple string prepeding in the example above should be natively changed
+    /// to some complex operation, as transactions may be tried multiple times.
```
spacejam, line 342: incorrect. transactions are serializable.

src/lib.rs
```diff
@@ -8,6 +8,11 @@
 //! are supported with the
 //! [`Db::open_tree`](struct.Db.html#method.open_tree) method.
 //!
+//! Sled is not designed to be used by multiple processes simultaneously.
+//! Although Sled tries to use advisory filesystem locking to
+//! protect the database from accidental access from foreign
+//! processes, this should not be reiled upon.
```
spacejam, line 14: no need for this comment here

src/db.rs
```diff
@@ -5,6 +5,9 @@ use crate::*;
 /// The `sled` embedded database! Implements
 /// `Deref<Target = sled::Tree>` to refer to
 /// a default keyspace / namespace / bucket.
+///
+/// Note that dropping `Db` does not gurantee closing the database. 
+/// Sled may still access the underlying directory under the hood.
```
spacejam, line 10: no need for this comment here

src/tree.rs
```diff
@@ -231,6 +231,8 @@ impl Tree {
 
     /// Perform a multi-key serializable transaction.
     ///
+    /// User-provided function `f` may be run multiple times in case of conflicts.
+    ///
```
spacejam, line 235: please get rid of all of the other comments except this one, which is the only one we discussed earlier.

src/lib.rs
vi: Where should it be instead?

I don't see other mentions of the shared multiprocess use cases in the documentation, yet questions do arise.

src/db.rs
vi: Where should it be (assuming the comment itself is not wrong, just misplaced)?

Filesystem usage aspect may be important for users, especially, for example, if the same process manages mounts/unmounts of the filesystem that backs the database.

src/tree.rs
vi: Shall I try to construct some concrete example (involving external I/O from within a transaction) where this break down?

spacejam:
Apologies for my harsh tone. I was feeling a bit frustrated at the feeling that I have been repeating the same thing, but better docs are a great way to address this frustration. I've reworded a few of your comments and placed them in places I hope will be visible to people who will encounter them. https://github.com/spacejam/sled/commit/66cf95758faac3cfbf837796c4cf35aa926f3ea3

src/lib.rs
spacejam: I've reworded this and placed it high up on the main README https://github.com/spacejam/sled/commit/66cf95758faac3cfbf837796c4cf35aa926f3ea3

PR #1238, opened 2021-01-06
Don't require Vec in sled::Db::import, just Iterator.

vi:
Without this change, all those first byte vectors (thing's types and names?) from the all trees are needed to be known in advance, before feeding items into the first tree.

This interferes with stream-importing things into the database from something other than `sled::Db::export` (e.g. over the network).

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -367,7 +367,7 @@ impl Db {
     /// ```
     pub fn import(
         &self,
-        export: Vec<(
+        export: impl IntoIterator<Item = (
             CollectionType,
             CollectionName,
             impl Iterator<Item = Vec<Vec<u8>>>,


spacejam:
This is not a general-purpose bulk-loading API, as it doesn't properly expose errors as they are encountered, but it just panics. It's really only intended to be used for database migrations, which are hopefully about to end in the next few weeks with 1.0, which will see the `export` method removed entirely, and only keep `import` for importing historical databases with a little less fuss than manual iteration.

vi:
Will there be alternative forward-compatible ways to serialize entire database to some simple structure, then restore it back (omitting calculate-able data such as indexes). Will naive "just iterate `sled::Db::tree_names`, then iterate everything in each tree" always produce a good copy of the database, even with some future features?

> This is not a general-purpose bulk-loading API ... only keep `import` for  ...

> /// for use in sled version upgrades.

Maybe documentation should be more explicit that it is aimed to be _only_ for migrations and custom usages are not supported?

PR #1244, opened 2021-01-10
Disable everything related to metrics if the feature is disabled

timokoesters:
From my measurements this decreases memory usage by more than 20 MB


spacejam:
Wow, good catch, thanks @timokoesters !

PR #1248, opened 2021-01-13
Use an in-memory overlay for receiving node updates before merging them into the backing NodeInner in a batch

spacejam:
This reduces the update pressure for contiguous nodes by buffering updates in an in-memory "overlay" that sits on top of an Arc-wrapped contiguous Node, and only when the node is rewritten due to GC or reaching the max frag threshold will it be consolidated in one go. 

This essentially maps the overlay to Link updates, and the backing NodeInner to fully rewritten nodes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,15 +18,17 @@
 
 ## Breaking Changes
 
-* #1214 The deprecated `Config::build` method has been removed.
+* #1135 The "no_metrics" anti-feature has been replaced with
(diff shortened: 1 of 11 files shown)


spacejam:
Thisis in-part inspired by reading the [SILT paper](https://www.cs.cmu.edu/~dga/papers/silt-sosp2011.pdf) where they store new items in a "LogStore" that then gets merged in bulk into a backing node that has a more compact representation.

PR #1261, opened 2021-01-15
Avoid allocations while processing mark_replace in SA

spacejam:


spacejam:
This clears all short-lived allocations in DHAT's "filter by short lived" view

Issue #1241, opened 2021-01-08
Lockup with multiple dbs awaiting on flush_async


koivunej:
1. **expected result**: all `Db::flush_async()` futures complete to either success or failure
1. **actual result**: with multiple `flush_async` polled near-simultaneusly futures are never woken up
1. **sled version**: 0.34.6, reproducer on `dfbb72dbcc8e47d040c7d57f6cf0e5fa644a5417`
1. **rustc version**: rustc 1.49.0 (e1884a8e3 2020-12-29), linkers tested: default, lld-9
1. **operating system**: Linux 5.4.0-58-generic #64-Ubuntu SMP Wed Dec 9 08:16:25 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux, kubuntu 20.04.1, some nvme drive, ryzen 9
1. **reproducer**: https://github.com/koivunej/sled/tree/reproduce_flush_async_lockup
1. **logs, panic messages, stack traces**: see below

Run the reproducer(s) with `cargo run --example repro`. Example reproducer output:

```
1: start
polled pending: 0x7ff52802ca38
polled pending: 0x7ff52402bd98
polled pending: 0x7ff53002c978
polled pending: 0x7ff53802c878
polled pending: 0x7ff58802bd98
polled pending: 0x7ff52c02bd98
polled pending: 0x7ff53402bd98
1: lockup
watcher completed as lockup happened as expected
---
process id: 468862
things should be steadily locked up now, entering sleep
```

The reproducer lockups every time on more than 7 core system at least (ryzen 9) but also hangs.

Observations:
- reproducers for tokio 0.2, 0.3 (see earlier commits in branch), 1.0 and async-std hint at this being a sled issue
- doesn't seem to happen with just `flush`
- making sure that the `db` outlives the `flush_async` future; no change

Looking the state with gdb once locked up:
 - tokio threads are idle awaiting for the wakeups
 - sled-thread-{0,1,2,3,4,5,6} ([stacktrace](https://gist.github.com/koivunej/2b7ce53d604abaf389f934db839727c5#file-sled-io-n) from first): https://github.com/spacejam/sled/blob/dfbb72dbcc8e47d040c7d57f6cf0e5fa644a5417/src/oneshot.rs#L56
[...]

PR #1263, opened 2021-01-15
Significantly simplify threadpool, and use different threads for diff…

spacejam:
…erent kinds of work to avoid dependency cycles.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -204,6 +204,7 @@ mod serialization;
 mod stack;
 mod subscriber;
 mod sys_limits;
+mod threadpool;
 pub mod transaction;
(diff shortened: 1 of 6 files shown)


spacejam:
@koivunej this may address your issue with flush_async in #1241. It shards different kinds of work across different threadpools to avoid dependency cycles where some blocking task is waiting on something that ends up scheduled later in its own work queue.

spacejam:
A nice discovery was just made: removing threadpool bounds causes memory usage to basically drop 95% on stress2 workloads :0

That means that almost all of our memory usage for stress2 up until now has been due to the threadpool work queue getting too long.

PR #1264, opened 2021-01-16
Import crossbeam-epoch and crossbeam-utils, simplify some things, get…

spacejam:
… rid of unused code, and perform dynamic collection when garbage grows large to better tame memory blow-ups when they happen.


spacejam:
* lower compile time due to inlining crossbeam-epoch and crossbeam-util and removing all unused functionality
* better recovery from latency spikes due to dynamic collection, preventing "garbage death spirals" that we previously sometimes witnessed

spacejam:
This also opens the doors to many optimizations over time, such as fine-grained collectors that have their memory usage tuned for specific data types, as well as better batching techniques for improving instruction cache hit rates while collecting garbage.

PR #1267, opened 2021-01-17
fixed-stride key omission

spacejam:


divergentdave:
I believe keys will always be monotonic, since they are stored in sorted order. Would "keys are an arithmetic sequence" be a better description for this?

spacejam:
fixed stride, linear, equally spaced, etc...

PR #1269, opened 2021-01-19
Add a specific truncation thread to make dependency loops between the…

spacejam:
… SA and blocking IO impossible


spacejam:
This addresses a deadlock that currently happens sometimes due to the SA blocking on the completion of a truncation operation that is queued up behind a stabilization operation that is in-turn blocked on the SA

PR #1273, opened 2021-01-21
feat: Enable the flusher for iOS

jesdazrez:
Hey there! Congrats on the awesome crate 🎉 .
While coding something on iOS we noticed the changes were not being synced.
From some commit messages I found that the rationale behind the cfg gate was to only enable the flusher in platforms with support for threads. This enables the flusher for iOS.

I can add some docs if that's a goal, just not sure where given that `Config`'s `flush_every_ms` options is `#[doc(hidden)]`.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -22,6 +22,7 @@ pub struct Context {
             target_os = "freebsd",
             target_os = "openbsd",
             target_os = "netbsd",
+            target_os = "ios",
         )
     ))]
     pub(crate) flusher: Arc<Mutex<Option<flusher::Flusher>>>,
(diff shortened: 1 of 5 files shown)


spacejam:
Hey @jesdazrez, thanks for adding this! I have just been conservative about what platforms receive the flush thread because it can cause more resources to be used sometimes, and I have preferred to enable them one at a time as users need them. Please open an issue if you notice that on iOS the flush thread is using a lot of resources in the gc function it performs!

jesdazrez:
I understand :). We will keep an eye open and report back if we find something odd 👍 .

PR #1283, opened 2021-01-26
Avoid removing and re-adding an item to the Lru upon cache access

spacejam:


spacejam:
This cuts ~5-10% of cpu cycles on some read-only workloads

PR #1304, opened 2021-03-04
[wip] incremental swizzle implementation

spacejam:
This is an incremental new approach of pointer swizzling, replacing #1300 which proved to be a bad approach that caused things to get bogged down. This PR will implement similar things over time, but with an eye toward keeping each commit more-or-less functional instead of the shotgun approach that didn't pan out in the previous attempt.

- [ ] first priority: paged-out single frag log items stored in a single pointer
  - [x] fuzzy snapshot
  - [ ] trim snapshot
  - [ ] make HeapId single-word
    - [ ] remove LSN from HeapId 
  - [ ] remove log pointers for heap items during snapshot creation
  - [ ] make DiskPtr single-word in standard case, moving to a single allocation while it's log + heap
  - [ ] add discriminant to low bits of pagetable entries
  - [x] remove CacheInfo dependencies from segment tracker
  - [ ] page-in fragmented pages on startup

- [ ] secondary goals
  - [ ] use swizzle discriminant + log coordinate for free entry not yet merged into snapshot
  - [ ] use null for free pages that have been merged into the last snapshot
  - [ ] collapse DiskPtr into inline enum with swizzled boxed coords for log+heap
  - [ ] make disk info inline with pages and frags to avoid the vec allocation
  - [ ] let frags replace single frags for the same key to avoid merges on hot keys
  - [ ] remove update enum from page representation

non-goals for this pr:
* intermediate node structs are ok for now


spacejam:
closing this due to it being completely replaced by the [marble-based approach](https://github.com/komora-io/marble) that should achieve better overall outcomes across the board

PR #1307, opened 2021-03-14
Add contains_tree method to the Db, related to issue 1297

mich2000:
Hello guys, this is my 1st contribution to the sled embedded database. This PR is to add a method to the db struct to be able to see if there is already a tree with that name in the database.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -470,6 +470,18 @@ impl Db {
     pub fn space_amplification(&self) -> Result<f64> {
         self.context.pagecache.space_amplification()
     }
+
+    /// Returns a true value if one of the tree names linked
+    /// to the database is found, if not a false will be returned.
+    pub fn contains_tree<V: AsRef<[u8]>>(&self, name: V) -> Result<bool> {
+        Ok(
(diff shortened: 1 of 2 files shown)


spacejam:
Thanks @mich2000 ! I'll make a small edit once I've merged this in to try to avoid iterating over the whole tree, and see if the key exists.

src/db.rs
```diff
@@ -470,6 +470,18 @@ impl Db {
     pub fn space_amplification(&self) -> Result<f64> {
         self.context.pagecache.space_amplification()
     }
+
+    /// Returns a true value if one of the tree names linked
+    /// to the database is found, if not a false will be returned.
+    pub fn contains_tree<V: AsRef<[u8]>>(&self, name: V) -> Result<bool> {
```
vilgotf, line 476: Isn't this `Result` unnecessary?

mich2000: Yeah you're right, I think I'm going to send out a new PR to work on that. I think it is because I literally copy-pasted the feature method.

Issue #1313, opened 2021-04-01
Subscriber as a Future occasionally skips records (but as an Iterator it does not) 


rrichardson:
1. Expected result:  That the inserted entries would exactly match the entries output by the Subscriber when poll()'d 
2. The Subscriber, when polled, seems to skip about 1.5% of records (when insertion runs in a tight loop) 
3. Version v0.34.6  - (from crates.io) 
4. Rust rustc 1.51.0 (2fd73fabe 2021-03-23) 
5. OS: Fedora CoreOS 
6. Reproduction:  https://github.com/rrichardson/sled_stream/blob/main/src/main.rs  
7. Hmm..  

```
Testing future
Future missed 13 records D:
Testing Iterator
Iterator's all good :)
``` 

I have not yet looked at the code, but I think I have an idea of what the problem might be if Iterator works and Future doesn't.. either way, as part of #1311 I am happy to submit a PR that changes the Future to a Stream and fixes this issue. 

(Also, it should probably documented that when Subscriber is used as an Iterator, it can block, this is certainly the intended behavior, but it might surprise people.)

rrichardson:
I added a repo for the new Batch scheme in the latest #main.   Basically the same result, except now stddev of missing entries is much wider. 

https://github.com/rrichardson/sled_stream/blob/batches/src/main.rs

rrichardson:
OK. In digging through the state machine of Subscriber and OneShot/Filler I see the problem: 

There are 2 scenarios basic scenarios  
1.  rx dequeues a oneshot, and that oneshot is already filled, this is the most likely scenario. 
2.  rx dequeues a oneshot, and that oneshot is not yet filled by the time it gets polled, this is less likely, but happens in tight loops

Scenario 1 work's fine. The oneshot is polled while filled, thus it returns the value from the fn. 

Scenario 2 is not fine, though:  The  `Poll::Pending` branch in the match is taken, so the function returns, thus dropping the OneShot receiver.  
However, the subscriber still has a valid waker, and the OneShotFiller is still in ReservedBroadcast.subscribers.  So it calls fill() and then the waker is hit. 
Then poll() is called again from the wake-up.  So it dequeues a *new* OneShot, and that one will behave correctly, because it hit Scenario 1.  
Sadly the previous oneshot is lost in time.  Like tears in rain. 

---- 

[...]

PR #1314, opened 2021-04-01
Store pending oneshot for subscriber future

asonix:
Fixes https://github.com/spacejam/sled/issues/1313

Add a field on Subscriber to store a oneshot after getting a pending poll. It's possible we'll need to pull some extra waker shenanigans in order to ensure we wake up when the oneshot becomes ready, but I'm not versed enough in this codebase to know that yet.
This change also removes an unneeded use of `unsafe` (I think this is fine, but let me know if it was there for a reason)

### DIFF
--- a/src/oneshot.rs
+++ b/src/oneshot.rs
@@ -63,8 +63,12 @@ impl<T> OneShot<T> {
     /// returning an error if not filled
     /// before a given timeout or if the
     /// system shuts down before then.
+    ///
+    /// Upon a successful receive, the
+    /// oneshot should be dropped, as it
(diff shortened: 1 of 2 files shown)


asonix:
It looks like OneShot is handling it's own waking, so this change should be fine

asonix:
I've implemented this fix for `next_timeout` as well, since previously it would drop incomplete oneshots.

spacejam:
Thanks for adding this fix, @asonix ! I am hoping to cut a new release this weekend if I can get a bit of necessary stabilization work done, which will include this fix :)

PR #1315, opened 2021-04-05
Feature/resolve tree contains

mich2000:
This PR is used to simplify the contains_tree to a simple bool instead of a result.

Edit note: This PR has also a cargo fmt applied to it, to make it more readable.


spacejam:
Thanks for the improvement, @mich2000 :) I'm hoping to do some stabilization work this weekend, and hopefully a new release will be cut soonish that includes the `contains_tree` method!

mich2000:
Nice man, feels good to contribute.

PR #1355, opened 2021-07-21
apply cargo clippy fix

mich2000:
By using cargo clippy on normal mode and with the testing feature on, I got rid of some cargo warnings that I got out of my own build and that of the pipeline.

The only warning I could not get through was on src/result.rs file on 84. So the fix was instead of cloning the bt variable, I just dereferenced it. But this made compiling with testing feature not work.

### DIFF
--- a/src/histogram.rs
+++ b/src/histogram.rs
@@ -104,7 +104,7 @@ impl Debug for Histogram {
 impl Histogram {
     /// Record a value.
     #[inline]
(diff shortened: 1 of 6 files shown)


mich2000:
Hey @spacejam do you want me to remove the indentations of my PR?

spacejam:
LGTM, thanks @mich2000!

PR #1359, opened 2021-08-20
Format, clippy suggestions

sigaloid:
Seems to pass tests

### DIFF
--- a/src/backoff.rs
+++ b/src/backoff.rs
@@ -6,12 +6,13 @@ const SPIN_LIMIT: u32 = 6;
 
 /// Performs exponential backoff in spin loops.
 ///
-/// Backing off in spin loops reduces contention and improves overall performance.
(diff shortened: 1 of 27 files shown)


spacejam:
Hello, could you please separate the clippy suggestions from the fmt suggestions? I don't want to needlessly merge formatting changes right now, as different people all have different rustfmt configs and this constantly seems to shift over time, but things that remove clippy warnings are generally OK to merge.

sigaloid:
Reverted

spacejam:
LGTM, thank you for cleaning up these small messy bits :)

spacejam:
upon closer look, I think the LRU code might have a bug in it, but no need to fix it. The main reformats from your split out PR have been merged. Closing this because the remaining changes that don't change semantics in possibly broken ways are just a small change from assert_eq to assert, so I'll do this in a separate branch real quick to get it moving.

PR #1360, opened 2021-09-04
transaction: adds a transaction_with_state method

baloo:
Transaction::transaction_with_state is intented to write data to an
outside state without making a copy.

This is useful to write to a buffer already allocated:
```rust
let mut state = State {
    buf: [0u8; 1024],
};
db.transaction_with_state(&mut state, |state, db| {
    if let Some(value) = db.get(b"k")? {
        state.buf[..value.len()].copy_from_slice(&value);
    }

   Ok(())
})?;
```

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -489,6 +489,79 @@ pub trait Transactional<E = ()> {
             }
         }
     }
+
+    /// Runs a transaction with a state, possibly retrying the passed-in
+    /// closure if a concurrent conflict is detected that would cause a
+    /// violation of serializability. This is the only trait method that
+    /// you're most likely to use directly.
+    /// ```
+    /// # use sled::{transaction::{TransactionResult, TransactionalTree}, Config};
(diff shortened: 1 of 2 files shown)


baloo:
Hi!
Thank you for sled! it's an amazing library.

This PR is merely a proposal for a problem I face, this would be my first contribution to sled.

Because the transaction closure is an `Fn()` and can be called multiple times it can not capture or write data from the outside. I'd like to avoid allocating data on the heap as this means extraneous copy I'd like to avoid in my use-case.

spacejam:
Hello @baloo, the reason why this is not safe is because transactions may be executed optimistically and repeat several times in the event of conflicts, causing any mutations to be incorrect unless they happen to be idempotent.

baloo:
@spacejam any suggestion how to achieve that? Beside allocating on the heap.

I fully understand the closure has to be idempotent, but in the example I shared I believe this is the case.

PR #1363, opened 2021-09-06
Additional clippy's from nightly

sigaloid:
image
Passes tests
Would have added these to #1359 but these were found with cargo +nightly, so if you dislike the style, feel free to close


spacejam:
LGTM :) Thank you for splitting this away from the larger text reformatting PR!

PR #1365, opened 2021-09-12
Cut 0.34.7 with Subscriber Future fix

spacejam:
This is a small point release, and a one-off small branch off of main, as work continues with a fairly large rewrite of the storage system before the next major release.

@uuhan


spacejam:
published

Issue #1368, opened 2021-09-15
impl futures::Stream<Event> for Subscriber


vi:
For some reason, `sled::Subscriber` implements `Iterator` and `Future` - a strange combination.

Instead of `Future`, `futures::Stream` would seem to be a more reasonable choice. How do I keep watching on more events asynchronously otherwise?

#### Use Case:

Watching multiple operations happening to a Sled database (like with `Iterator` impl), but asynchronously.

#### Proposed Change:

```rust
impl futures::Stream for Subscriber {
    type Item = Event;
    ....
}
```

#### Who Benefits From The Change(s)?

Async Sled users.

#### Alternative Approaches

* Spawning blocking task to reach the `Iterator` impl
* Repeat `watch_prefix` after each received event. Assign sequence number to each entry in the database and explicitly retrieve entries that were added to the database while we were not "on the watch".

asonix:
As noted in the documentation, you can use
```rust
while let Some(item) = (&mut subscriber).await {
}
```
But you're right that a `.next()` method or similar would be more intuitive

vi:
Typically resolved `Future`s should not be polled again:

> Once a future has finished, clients should not poll it again.

Relying on reusing Future after non-`Pending` poll is like relying on `Iterator` becoming live again after already returning `None` once. Even if it works, it does not play nice with various combinators and hinders understanding the code.

`futures::Stream` is a canonical "multi-use Future".

busyboredom:
This feature would also be very nice for Actix users. I had a hell of a time trying to use a Subscriber as a second stream for an actix websocket (I ended up just giving in and polling instead).

vi:
@busyboredom , I expect that one can workaround it with a manual `Stream` implementation where `poll_next` delegates to Sled subscriber's `poll`.

tv42:
This sounds like a duplicate of https://github.com/spacejam/sled/issues/1311

The reworking of the idea to a sink sounds nice to me.

PR #1372, opened 2021-09-16
Delegate Future implementation to Next struct

asonix:
Inspired by https://github.com/spacejam/sled/issues/1368

I don't know if this is something you care to merge or not, but I've done the work to implement a more "standard" way to interact with Subscriber asynchronously

Previously, reading values out of Subscriber asynchronously would involve awaiting an `&mut Subscriber`, which causes re-polling the same future after it yields a value. This is considered 'improper' for a Future, and more closely maps to the Stream  trait. Unfortunately, `Stream` has not been standardized and so implementing that trait would require pulling in the `futures-core`   dependency, which is not yet 1.0 and subject to change. Instead, this implements a Stream-like `poll_next` method on  `Subscriber`, and exposes that from an implementation of Future for a new `Next<'_>` type, which can be constructed by calling `.next()` on a Subscriber. This allows for a more traditional interaction with futures that mimics how other libraries produce values.

Note that it is still possible to indefinitely poll on an `&mut Next<'_>` type to pull every value out of the subscriber, but it is less intuitive to perform that operation and it likely won't occur in downstream projects.

### DIFF
--- a/src/subscriber.rs
+++ b/src/subscriber.rs
@@ -107,9 +107,9 @@ type Senders = Map<usize, (Option<Waker>, SyncSender<OneShot<Option<Event>>>)>;
 /// ```
 /// Aynchronous, non-blocking subscriber:
 ///
-/// `Subscription` implements `Future<Output=Option<Event>>`.
+/// `Subscription` provides a `next` method which returns an `impl Future<Output=Option<Event>>`.
 ///
-/// `while let Some(event) = (&mut subscriber).await { /* use it */ }`
+/// `while let Some(event) = subscriber.next().await { /* use it */ }`
 pub struct Subscriber {
     id: usize,
     rx: Receiver<OneShot<Option<Event>>>,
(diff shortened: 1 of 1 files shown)


src/subscriber.rs
```diff
@@ -125,6 +125,44 @@ impl Drop for Subscriber {
...
+    /// Creates a future that resolves to the next value of the
+    /// subscriber, or None if the backing `Db` shuts down
+    pub fn next(&mut self) -> impl Future<Output = Option<Event>> + '_ {
+        Next { subscriber: self }
+    }
+
+    fn poll_next(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Event>> {
```
asonix, line 137: This API matches the current definition of the `Stream` trait in Futures 0.3, so if that ends up stabilizing as-is then a Stream implementation will be "free"

Otherwise some changes will need to be made later down the line, but it's a private method anyway

src/subscriber.rs
```diff
@@ -125,6 +125,44 @@ impl Drop for Subscriber {
 }
 
 impl Subscriber {
+    /// Creates a future that resolves to the next value of the
+    /// subscriber, or None if the backing `Db` shuts down
+    pub fn next(&mut self) -> impl Future<Output = Option<Event>> + '_ {
```
asonix, line 130: Although this is returning a concrete Next<'_> type in the implementation, we don't need to leak the new type in the public API in the event it needs to change later for some reason

src/subscriber.rs
```diff
@@ -216,6 +218,23 @@ impl Iterator for Subscriber {
...
+}
+
+impl<'a> Future for Next<'a> {
+    type Output = Option<Event>;
+
+    fn poll(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Self::Output> {
+        Pin::new(&mut *self.subscriber).poll_next(cx)
```
asonix, line 234: `self.subscriber` is an `&mut Subscriber` but it can't move, so `&mut *self.subscriber` to make it move-able.

src/subscriber.rs
```diff
@@ -239,10 +258,7 @@ impl Drop for Subscribers {
 }
 
 impl Subscribers {
-    pub(crate) fn register(
-        &self,
-        prefix: &[u8]
-    ) -> Subscriber {
+    pub(crate) fn register(&self, prefix: &[u8]) -> Subscriber {
```
asonix, line 261: Unintentional formatting change. I can put this back if you want

src/subscriber.rs
asonix: Realizing now that `next` here conflicts with Iterator's `next` Maybe naming this something else would be better

Issue #1306, opened 2021-03-11
Implement From<&[u8; N]> and From<[u8]> for all N for IVec


xa888s:
# Why
Currently one can only convert arrays up to a certain limit (32) to an IVec, with this change one can convert from arbitrary arrays.
```rust
fn main() {
     let arr = [0, 1, 2, 3];
     let iv = IVec::from(&arr);
}
```
This also means there will be one less macro, which is usually a good thing.

# Why not
This would push the MSRV up to 1.51.0 (which as of writing is in beta), which would be a substantial change. There is also the point that the macro satisfies many uses of the From implementation, and this would be an API change that cannot be reversed without breaking compatibility.

# When
Probably not until some other more substantial feature arises that warrants the MSRV bump.

Spaceface16518:
[Rust 1.51.0 was release on March 25, 2021](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html), allowing this change on stable.

spacejam:
I would like this change eventually. Historically I've tended to bump the MSRV pretty conservatively after feeling kind of swamped by people telling me things were broken after I cut releases based on new stable versions, and in my mind, something like trailing by 6 months feels appropriate to reduce the number of issues that will be opened by people with old rustc versions.

mich2000:
A PR for this issue has been made on https://github.com/spacejam/sled/pull/1376.

PR #1376, opened 2021-10-03
Chore/ivec const usize

mich2000:
Because we are now over 6 months of the msrv 1.51 implementing const generics, this PR will aim at simplifying the from trait implementation of the IVec struct.


mich2000:
This PR is linked to https://github.com/spacejam/sled/issues/1306.

spacejam:
hi @mich2000, thanks for the nice clean-up! I have actually just merged another PR that does the same thing, only because it was the first one that I opened, but they are both identical I think.

Issue #1380, opened 2021-10-12
A quite serious bug when use Subscribers as a Future


uuhan:
Finally I can reproduce the blocking issue. (Related issue: #1378 )

code sample: 
```toml
[dependencies]
sled = "0.34.7"
tokio = { version = "1.12.0", features = ["macros","rt-multi-thread", "time"] }
```
```rust
#[tokio::main(flavor="multi_thread")]
async fn main() {
    let db = sled::Config::new().path("test.db").open().unwrap();
    let tree = db.open_tree(b"tree").unwrap();

    let mut subs = tree.watch_prefix("");

    tokio::task::spawn(async move {
        for i in 1usize..2024 {
            println!("{}", i); // this can sometimes reaches 1025, then blocks
            tree.insert(&i.to_le_bytes(), &i.to_le_bytes()).unwrap();
            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
        }
    });

    tokio::task::spawn(async move {
        println!("watch");
        while let Some(event) = (&mut subs).await {
            match event {
                sled::Event::Insert { key, value } => {
                    println!("event: {:?} {:?}", key, value);
                }
                _ => {}
            }
        }
    });

    tokio::time::sleep(std::time::Duration::from_secs(10000)).await;

}

```

Run at least twice

```shell
cargo run
```

1. expected result
    Subscribers should work as expected.
2. actual result
    Not work as expected
3. sled version
   0.34.7
4. rustc version
    rustc 1.57.0-nightly (97032a6df 2021-09-08)
5. operating system
   NAME="Ubuntu"
   VERSION="20.04.3 LTS (Focal Fossa)"
    and
   MacOs Big Sur M1
6. minimal code sample that helps to reproduce the issue
    as above
7. logs, panic messages, stack traces

uuhan:
fixed by #1381 🥳

PR #1381, opened 2021-10-13
Fix: insert deadlock #1380

uuhan:
After 1024 insertions of the same key-value pair, the SyncSender blocks.

This fixes issue #1380


spacejam:
thanks for this fix, @uuhan !

PR #1383, opened 2021-10-16
Use const generics for IVec::from array

mbartlett21:
This will move the minimum supported rust version to 1.51, which may or may not be acceptable.


spacejam:
@mbartlett21 nice clean-up, thanks! 1.51 is totally acceptable at this point IMO

mbartlett21:
Thankyou :)

PR #1385, opened 2021-10-21
Fixes 32 bits systems compatibility

Tpt:
The build was failing when `size_of::<usize> < 8` with the error "this arithmetic operation will overflow".

This error is not emitted when running "cargo check" and hence not caught by the CI.


spacejam:
thanks :)

PR #1391, opened 2021-12-22
Docs: add example for a global lazy_static instance

PAN-Ziyue:


spacejam:
hi @PAN-Ziyue, thank you for taking the effort to add these docs, but I am going to close this issue because there is a good chance that the next released version of sled will not work the same way with globals. Specifically, I'm probably going to make Tree/Db !Sync, for the reason that this will force users to clone Tree/Db and give each accessing thread their own handle. This will allow certain functionality around epoch based reclamation to be about twice as fast as what is currently the case using the crossbeam_epoch crate, because we're able to avoid more memory barriers and other overhead if we assume that per-thread garbage bags are not accessed by other threads.

leocornelius:
Sorry to comment on a closed issue. I am wondering about the best way to use sled in multiprocess contexts. I am thinking of making a "wrapper" struct arround a sled db object like so:
```rust
struct Database {
    pub masterdb: SledDb
....
```
This wrapper struct would implment a few functions like read_value(tree_name, key), write_value(tree_name, key, value) etc. THe entry point of the program (the cli) creates this object and sets the masterdb object (eg the single db which tress are opened on). Then for each different aspect of the program (eg the network manager, the api, etc) this object is cloned and passed to the initialiser of the manager of that aspect (eg P2pManager, ApiManager, SyncManager, etc). When the cli exits, it would send an exit signal to all the managers, who then drop all their references etc. This leaves a single Database struct (owned by the cli) which i can then call .flush() on and drop. Is this an ergonomic and good approach for using sled in a multicomponet/process context? Another question would be if the diffrent coppies of the sledDb object (the one contained in the masterdb varible) would be "synced". Eg if i call open_tree on it on one coppy and write some values, would those changes reflect on the other coppys (so if i opened the tree and read those values they woild be up to date). Thanks

leocornelius:
I will open a new issue for this so it can be seen

PR #1409, opened 2022-04-26
Ensure proper file permissions on unix

jsparber:
The database shouldn't be world readable.


vi:
Shouldn't this be decided by `umask` by default?

If user does not want to have information world-readable by default, just set umask `077`.

spacejam:
I agree with @vi - this is a concern that should be handled at the umask level

PR #1418, opened 2022-09-11
Prevent UB with `assume_init` on uninit bytes

clubby789:
If `size_of::<F>` < `size_of::<Data>`, the remaining bytes would be uninitialised - thus, calling `assume_init` would cause UB, as noted by Miri. Storing the data as `MaybeUninit` and casting to the correct `FnOnce()` object before reading it fixes this, only using initialized bytes.


spacejam:
Thanks :)

clubby789:
This seems to be failing on a lot of unrelated Clippy lints 🤔 Auto applying them results in a diff of 22 files with more not automatically fixed, so this probably isn't the PR to address these in.

PR #1421, opened 2022-09-30
Apply `rustfmt` and `clippy` fixes across crate

clubby789:
Add clippy and formatting fixes to allow CI to pass (#1418).


clubby789:
This seems to cause a few tests
image
to fail locally, possibly the automatic clippy fixes broke the code.

clubby789:
I also disabled the `missing_const_for_fn` lint as that has some issues with false positives (i.e. when an operation is only `const` with unstable features)

spacejam:
this would cause too many conflicts with private branches. If there are specific correctness-related diffs, please feel free to open a separate PR with them.

PR #1426, opened 2022-10-02
Remove zstd in anticipation for it being re-added through marble

spacejam:
zstd should not be used on the write-ahead log, but only on large blocks of items with a custom dictionary, which is provided by the [marble](https://github.com/komora-io/marble) crate that sled's storage is being restructured on top of.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 default.sled
+crash_*
 *db
 *conf
 *snap.*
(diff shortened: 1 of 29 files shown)


flokli:
I got confused quite a bit by the state of the tree currently, and opened https://github.com/spacejam/sled/issues/1491. It'd be good if we could add a disclaimer, so people know about what to expect of the current tree.

PR #1432, opened 2022-11-29
Fix typos

kianmeng:
Found via `typos --format brief`


BartTerpstra (approved):
Seems to just be spelling fixes in comments and warning messages. Unless something is doing string comparison on the warnings and errors (💀), this will break nothing.

spacejam:
I think I need to add that tool as a pre-commit hook in my local setup - thanks :)

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #1435, opened 2022-12-08
Improve `Config` docs: Document defaults

alexheretic:
Hello, I started using _sled_ in [ab-av1](https://github.com/alexheretic/ab-av1/pull/92) & it seems to work well for me thank you!

Having the default config values in documentation would have sped up my onboarding, saving me from searching in the source for the values. So this PR adds that info to the docs.


alexheretic:
Any interest in this? If not I'll close to avoid noise.

PR #1446, opened 2023-06-26
structured data support

Walker-00:
Structured data type like struct, enum and others support added under the future "json" tested for window and linux I'm not sure about mac It's complicated for me to cross compile for it.

### DIFF
--- a/src/json.rs
+++ b/src/json.rs
@@ -1,0 +1,44 @@
+//! structured data store support for `sled` db
+//! # Examples
+//!
+//! ```
+//! use serde::{Deserialize, Serialize};
+
+//! #[derive(Serialize, Deserialize, Debug)]
+//! struct Info {
+//!     name: String,
+//!     age: u8,
+//! }
+
+//! let linus_walker = Info { name: "Linus Walker".into(), age: 14 };
+//! let db = crate::open("db").unwrap();
+//! db.insert_json("linus_walker", &linus_walker).unwrap();
+//! let resul = db.get_structued::<Info>(&"linus_walker".to_string()).unwrap();
+//! println!("{resul:?}");
+//!}
+//! ```
+
+use crate::{Db, IVec, Result};
+use serde::{de::Deserialize, ser::Serialize};
+
+impl Db {
+    /// function implementation for `Db` to store structured data as json in db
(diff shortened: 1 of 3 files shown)


src/json.rs
```diff
@@ -0,0 +1,55 @@
...
+use serde::{de::Deserialize, ser::Serialize};
+
+impl Db {
+    /// function implementation for `Db` to store structured data as json in db
+    pub fn insert_json<K: AsRef<[u8]>, V: ?Sized + Serialize>(
+        &self,
+        key: K,
+        value: &V,
+    ) -> Result<Option<IVec>> {
+        let value = serde_json::to_string(value).unwrap();
```
dariusc93, line 32: Would be best not to unwrapping as that would result in a panic if it fails to serialize (you could probably use io::Error or just one of the other errors in `result::Error`). Also, wouldnt it be better to use `serde_json::to_vec` instead?

src/json.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /// function implementation to get structured value as output from database
+    pub fn get_structued<T: for<'a> DeDeserialize<'a>>(
+        &self,
+        key: &dyn AsRef<[u8]>,
+    ) -> Result<Option<T>> {
+        match self.db.get(key.as_ref())? {
+            Some(v) => {
+                let x = String::from_utf8(v.to_vec()).unwrap();
```
dariusc93, line 43: You could grab the bytes and pass it to `serde_json::from_slice` instead

dariusc93 (commented):
Left some comments. 

Although in my case, I would generally do this as a wrapper around sled for storing serializable objects, but what could be done to make it more uniform for other implementations around serde (eg bincode) would probably be to have 2 function pointer stored in `Db` that would call eg `serde_json::to_vec` and `serde_json::from_slice` (or in bincode case,`bincode::serialize` and `bincode::deserialize`) and have the functions be something like `Db::insert_ser` and `Db::get_de` (not the exact naming, but just throwing something out there) for anything related to serde under say a `serde` feature.

spacejam:
Hello - thank you for your interest in this, but I am going to close this PR because I would like sled to be completely ignorant of higher level formatting concerns for the forseeable future.

PR #1449, opened 2023-07-14
Wrong atomic ordering in clone

wang384670111:
https://github.com/spacejam/sled/blob/69294e59c718289ab3cb6bd03ac3b9e1e072a1e7/src/ivec.rs#L24-L29
https://github.com/spacejam/sled/blob/69294e59c718289ab3cb6bd03ac3b9e1e072a1e7/src/ivec.rs#L33-L50
https://github.com/spacejam/sled/blob/69294e59c718289ab3cb6bd03ac3b9e1e072a1e7/src/ivec.rs#L26
Use of Ordering::Relaxed is incorrect. The compiler is free to reorder, which may cause the result of the IVec(self.0) in thread 1 to be stored in hardware before the execution rc.fetch_add(suppose the clone function is executed in thread 1), if rc is 1 and the drop function is executed in thread 2, rc is judged to be 0 (at this point there is no execution rc.fetch_add in thread 1), which will cause the IVec (self.0) returned by thread 1 to have been freed . It should be Ordering::Acquire instead.

### DIFF
--- a/src/ivec.rs
+++ b/src/ivec.rs
@@ -23,7 +23,7 @@ pub struct IVec([u8; SZ]);
 impl Clone for IVec {
     fn clone(&self) -> IVec {
         if !self.is_inline() {
-            self.deref_header().rc.fetch_add(1, Ordering::Relaxed);
+            self.deref_header().rc.fetch_add(1, Ordering::Acquire);
         }
         IVec(self.0)
     }


spacejam:
Hey @wang384670111, thank you for looking into this! This is also something that seemed dangerous to me at first when I first found the general pattern of atomic orderings in the std::sync::Arc implementation, which this was ported from. The comments for [Arc::clone](https://doc.rust-lang.org/src/alloc/sync.rs.html#1503-1513) (referencing [the boost documentation talking about how reference counters can be incremented with Relaxed](https://www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html)) and the comment in [Arc::drop](https://doc.rust-lang.org/src/alloc/sync.rs.html#1861-1900) are all relevant here. I think the specific situation you describe would count as a use-after-free bug due to an owning thread (t2) executing drop on an IVec that only had a reference count of 1, while t1 was concurrently accessing it. Please feel free to re-open this if you think that the situation with IVec has an important difference with the above links, as I might have missed something and would be grateful to hear about an incorrect assumption here. Especially because I've copied this same pattern into the [`inline-array`](https://github.com/komora-io/inline-array) crate as part of sled 1.0.

wang384670111:
I took a closer look at the code in Arc::clone and Arc::d rop, and I think the situation described is the same as IVec, like the documentation says that the original reference prevents other threads from erroneously deleting the object. I think it's OK to use Relaxed, but I've studied IVec carefully, and some of the ordering in it is still a bit misused (using too strict SeqCst, etc.), and I'll submit some prs to discuss later.

PR #1453, opened 2023-07-28
Rename "testing" feature to "for-internal-testing-only"

joshtriplett:
Rather than having a feature named "testing" that external users might
be tempted to incorrectly use (and a detailed comment discouraging them,
which people may or may not see), rename the feature to
"for-internal-testing-only" so that nobody gets any wrong ideas about
whether it's meant for them.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,7 @@ jobs:
     - name: cargo test
       run: |
         rustup update --no-self-update
-        cargo test --release --no-default-features --features=testing -- --nocapture
+        cargo test --release --no-default-features --features=for-internal-testing-only -- --nocapture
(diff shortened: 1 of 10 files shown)


joshtriplett:
Sorry, I sent this for the `main` branch before realizing that sled 1.0 development is on a different branch. Doesn't look like this is needed on the latter. If you're not looking for changes on the `main` branch, feel free to close this.

spacejam:
Thanks, Josh :) While the 1.0 branch doesn't have this yet, it will receive some amount of test-only event tracking / expensive assertion verifiers that will use a feature, and I like using the feature name you proposed, so I'll merge it into main and port it to 1.0 when adding the new verification layers.

PR #1454, opened 2023-07-29
bloodstone: A few nits

wackbyte:
Fixes a few clippy lints and makes it so that `as_ref` is only called once on keys (out of concern for impure `AsRef` implementations).


spacejam:
thank you! I'll try to release a new alpha with your changes today

PR #1485, opened 2023-11-11
fix(benchmarks): fixed criterion benchmark not compiling on windows

Wicpar:
fixed criterion benchmark not compiling on windows due to jemalloc


Wicpar:
needs to be compiled on linux to be sure before merging

PR #1492, opened 2023-12-15
Lossy signature for Tree::set_merge_operator

RuofengX:
# Motivation

I am writting a double-index ecs component storage, by wrap the sled api. 

```rust
// Define a merge function trait like MergeOperator, let my lib user to write the function
pub trait MergeFn: Fn(EID, Option<Value>, Delta) -> Option<Value> + Send + Sync{} 
```
I found that it is not possible to  insert a trait-object(`Rc<dyn MergeFn>` in my case) into a Tree as its merge_operator. Because the set_merge_operator function's signature is strict to a static implement.

I design a trait to let my user write there own closure, an wrap them into a `Rc<>`. Function pointer is also a workround but use triat object is encouraged by the Book, see more in [here](https://doc.rust-lang.org/book/ch19-05-advanced-functions-and-closures.html#function-pointers).

# PR content

- Just change the signature to a Box<...>.   
- All test cases related are changed.  
- `cargo test --features light_testing` are passed.

# 1 more thing
(Not include in this pr)

In my scanery, merge function is considered as `stateless`, so it's a `Fn` trait, instead `FnMut`, also the merge function is init at the beginning of the program, and never changed after that. So I think the mutex to protect the function isn't needed.  The merge function should be passed at the init pharse of the Tree when the program begin, and not protected by a Mutex.

## Pros

- No cost when set or call the merge function during run-time.
 
## Cons

- One tree can only hold one immutable merge function during its lifetime.
- User must impl their own inner-mutable data in the merge function.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "rust-analyzer.cargo.features": [
-        "light_testing"
-    ]
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "rust-analyzer.cargo.features": [
+        "light_testing"
+    ]
+}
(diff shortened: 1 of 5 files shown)


RuofengX:
It change the public api.

PR #1494, opened 2023-12-27
Added xtensa architecture to cfg for atomic_shim

gpgreen:
I am building an ereader project based on an xtensa board. I would like to use sled as a db, but it doesn't compile with the stable version 0.34.7.

xtensa architecture doesn't have 64bit atomics, so adjusted config to bring the shim versions of those types

### DIFF
--- a/src/atomic_shim.rs
+++ b/src/atomic_shim.rs
@@ -3,10 +3,16 @@
 #[cfg(not(any(
     target_arch = "mips",
     target_arch = "powerpc",
+    target_arch = "xtensa",
     feature = "mutex"
 )))]
 pub use std::sync::atomic::{AtomicI64, AtomicU64};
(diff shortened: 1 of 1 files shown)


black-puppydog:
@gpgreen it's been exactly one year but if you do remember: was tis patch the only thing you needed? I just tried compiling with it and it still came up with "type Atomic* does not exist" kind of errors.

gpgreen:
I haven't worked on the project for a while, and it's a complicated project, with no deadlines. I don't think I'll be able to answer quickly right now. xtensa was a very active project so it's quite possible there is another path to getting sled to work now. If you don't want to dig into it, just close, and I'll try again later when I get back to it.

black-puppydog:
@gpgreen no don't worry, I'm not a maintainer, just someone also trying to compile sled into an esp32 project at the moment. 😆

gpgreen:
@black-puppydog You had the "mutex" feature turned on?

PR #1516, opened 2025-01-22
Fix indentation in markdown list

TannerRogalsky:
We missed a newline in this list which cases the markdown list to be rendered incorrectly. Thanks for this project!


spacejam (approved):
Thanks :)

PR #1537, opened 2026-03-26
Fix stale snapshot removal log message

cuiweixie:
## Summary

Corrects the warning when pruning a stale snapshot file: the message previously printed the same snapshot id twice because both placeholders used `{id}`. It now reports the removed stale snapshot (`snap_id`) and the superseding snapshot (`id`).

## Context

Small correctness fix for log output in `enumerate_logs_and_snapshot` (metadata store).


spacejam (approved):
Thank you!

